PR #1603, opened 2016-02-20
Create issue_template.md

Aokromes:
Add issue template when creating new reports.


issue_template.md
```diff
@@ -0,0 +1,20 @@
+**Description**:
+
```
a-ladanov: Реклама рмн sports.ru

PR #3699, opened 2016-12-09
Disable stealth mode on door2windows.com

Bluscream:
Disable referer manipulation to allow downloads on door2windows without excepting the whole page


Alex-302:
Added manually.

Bluscream:
:+1:

PR #4454, opened 2017-02-24
added service push4site.com

getnaked:


Alex-302:
why not `||push4site.com^$third-party` ?

getnaked:
I'm sorry, you're right - the correct version ||push4site.com^$third-party

Alex-302:
done

PR #6616, opened 2017-09-05
Update specific.txt - rarbg.to

zero77:
Example URL: rarbg.to/torrent/tnvz4r2 and any other torrent enterys.
This only works for their main domain not proxies.


Alex-302:
Screenshot, please.

zero77:
I have attached two screenshots of be for and after the rule was used.

Also because of the “rarbg.to##” part it will not work with proxy domains like: rarbg-to.proxydude.xyz and rarbg.unblocked.lol, should I remove the “.to” part or use wild card.

Before:
before


After:
after

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/new
I can't add your rule.

PR #6633, opened 2017-09-07
Update specific.txt - deepdotweb.com

zero77:
Blocks annoying adds on the side of the page (both top and bottom).

Example URL:
deepdotweb.com


Alex-302:
Please use this button
https://github.com/AdguardTeam/AdguardFilters/issues/new

PR #6972, opened 2017-10-02
Blocking Web Browser CPU Mining

norbertjoni:


vitalkanev:
Some of the lists are binary files, so I can't read it properly

noorbertt:
https://github.com/AdguardTeam/AdguardFilters/commit/ca2d4c0c85f7dd2cd3d6bda57df8eda541364090
Now?

Alex-302:
Please don't make pull requests. Create an issues instead.
In general, we don't block coinhive. We show notification about it, and user can allow it or forbid.

PR #7159, opened 2017-10-20
Move Olark to "widgets" list its Chat section

aawilson:
Categorizing this more specifically according to its actual purpose.

People who are looking to block "spyware" (a huge list) might not want to block chat products in the process (a much smaller list). Or vice versa, honestly.

Full disclosure, I work for Olark, so infer whatever nefarious motives you need to from there. For what it's worth, I completely sympathize with people who want to block us, I just want to make sure that people aren't blocking us as a side-effect of blocking something unrelated (because that causes support load for us).

### DIFF
--- a/AnnoyancesFilter/sections/widgets.txt
+++ b/AnnoyancesFilter/sections/widgets.txt
@@ -139,11 +139,15 @@ javtasty.com##.table > .block-chat
 anime-skorpik.com###sgeChat
 taraftarium24-tv.com###free-chat-nandox
 tuttosportweb.com###sidebar
+||assets.olark.com^
 ||chatbro.com/embed_chats/
+||log.olark.com^
 ||nandox.com/*free-chat^$domain=taraftarium24-tv.com
+||nrpc.olark.com^
 ||secure.livechatinc.com^$third-party
 ||sge-chat.skorpik.ru^
 ||smartsuppchat.com/loader.js
+||stats.olark.com^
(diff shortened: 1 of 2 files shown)


Alex-302:
Hi
Could you, please, provide example of the page with this widget?

aawilson:
Absolutely:

https://www.olark.com/
https://www.coffeecommitment.com/
https://www.premiumecigarette.com/
http://rutherfordcountytn.gov/countyclerk/

I can get many more examples if you need them.

Alex-302:
Hmm... How can I see this widget? Can't find.

aawilson:
It's typically a floating element in a corner, an expandable box that becomes a live-chat interface.

image
image
image

Alex-302:
Ok. I understand now.
Moved to annoyances filter:
```
||nrpc.olark.com^
```

aawilson:
Your commits almost get there, but you'll also have to move `assets.olark.com`, and you'll also need to allow `/a` from `(static|assets).olark.com` (`/js`, `/app` and `/modules` might also need to be added, but I don't remember offhand which of those are still actually used and which are just legacy). `assets`'s content is identical to `static`, but one bypasses CDN (used to resolve caching issues). Keeping log and stats on the other list is probably fine (they're both for different aspects of tracking aggregate loader problems, so having a few people opt out of that won't kill us).

Alex-302:
Removed
`||assets.olark.com^`

PR #7624, opened 2017-11-14
Annoyances: silence intercom.com push notification

bpartridge:
Sites like www.alphaflow.com (as of 2017-11-14T20:52Z, see in an incognito window) and any others using Intercom's Engage product will pop up a loud autoplaying audio notification on the load of their widget. The widget and the message themselves are useful (as this is usually used for B2B product support) but the audio is incredibly disruptive (especially in work environments).


Alex-302:
Hi
I need example of the page with this widget. We can block in in annoyances filter.

ameshkov:
@Alex-302 
https://www.alphaflow.com/ is that example.

Moreover, shouldn't we block intercom completely in the annoyances filter? It falls under the filter policy.

Alex-302:
Added
```
||widget.intercom.io^$third-party
```

PR #7630, opened 2017-11-15
Update tracking_servers.txt

getnaked:
Add api.flocktory.com
The service tracks users and sends promotional emails.


Alex-302:
added

PR #10140, opened 2018-01-20
fix #6945

### LINKED ISSUE #6945: moviechat.org - Missed ad
https://moviechat.org

screen shot 2017-09-28 at 4 58 38 am

</details><br/>

***System configuration***

Information                         | Value
---                                 | ---
Operating system:                   | Android
Browser:                            | Chrome
AdGuard version:                    | 2.9
Filters enabled:                    | 1 2 3 4 14 11
AdGuard mode (Android only):        | VPN
Filtering quality (Android only):   | High-quality 
HTTPS filtering (Android only):     | On (Whitelist)
Helpdesk ID:            | 1635871

anewuser:
They're now using `https://moviechat.org/images/aff/` for image ads, and the CSS rule just removed a few blank pixels. It's better not to hide the section for it to be possible to know if they've changed the image URLs again.

### DIFF
--- a/EnglishFilter/sections/specific.txt
+++ b/EnglishFilter/sections/specific.txt
@@ -1408,8 +1408,7 @@ convertinmp4.com###b728l
 primewire.ag##a[href^="/gohere.php?"][style*="images/download_button_hover"][class]
 cs-fundamentals.com###right-col-block-ad
 photopea.com##.rightmost
-||moviechat.org/images/affiliates/
-moviechat.org##.adc
+||moviechat.org/images/aff
 theage.com.au##iframe[name="compare_save"]
 theage.com.au###jobs
 turbobit.net##a[href*="&campaign="] > img


Alex-302:
Please check again. @AdamWr added new rules.

anewuser:
Doesn't Adguard support ``||moviechat.org/images/aff``? That's what I had submitted to delete `.../affiliates/`, but instead he added `.../aff/` and kept the redundant `../affiliates/`.

He's also added a new redundant CSS rule instead of removing the existing one. As I said, the section is blank and just one or two pixels tall. There's no point in having any CSS rule to hide it.

AdamWr:
Hi,
Ok, I removed these rules
`||moviechat.org/images/aff/`
`||moviechat.org/images/affiliates/`
and added this - `||moviechat.org/images/aff`

Regarding CSS rule (`moviechat.org##.adc`) I removed it too, because the same rule is in EasyList, but I don't agree that this space shouldn't be hidden and from what I know, users usually want to hide this type of ad leftovers.
From what I see it has more then two pixels.

image

</details>
<br>

I added that rule:
`moviechat.org#$#.adc { position: absolute!important; left: -3000px!important; }`
because they use inline style `display: block !important;` which can override AdGuard element hidding rules.

PR #11201, opened 2018-02-07
Update foreign.txt

diomed:


Alex-302:
I need explanations for the rules/

diomed:
What explanation do you need?
These are all ad service domains, and last one is a government site 
that asks of people to open modal for login, so it has to be allowed.

Alex-302:
contacts-store.com is not working.
Please add `^$third-party` to domain blocking rules and comment to the last rule, why it must be completely excluded.

Alex-302 (changes requested):
contacts-store.com is not working.
Please add ^$third-party to domain blocking rules and comment to the last rule, why it must be completely excluded.

diomed:
I don't understand last part of your sentence. What comment should I add, and how is it completely excluded? I basically want that modal to open up. Can you please edit file instead of me?

Alex-302:
> What comment should I add,

we always add comments for exclusion rules with the reason of excluding.
As I see, nothing is broken on this site.

Alex-302 (approved):

Alex-302:
thanks, merged

PR #11814, opened 2018-02-18
metabans.com - site doesn't exist anymore

SpicySpaceman:
That site doesn't exist anymore, so the rule can be removed.


danidoedel:
At least you could explain why...

Alex-302:
damn. I thought Blaz removed it)

PR #15717, opened 2018-05-05
Update RussianFilter/sections/adservers.txt

davidhedlund:
Added three new servers.


davidhedlund:
Is is ad server-related filters for sites that are popular in Russia.

Alex-302:
Where is it used?
And we don't need social rules in Russian filter.

Alex-302:
Added `||ad.iplayer.org^`

PR #16100, opened 2018-05-12
bbc.co.uk - false positive

Bo98:
Example: https://www.bbc.co.uk/programmes/b0070hvg


Alex-302:
I can't reproduce it.
Please send report or create new issue.

Bo98:
@Alex-302 Oh I just noticed it might be the `$~third-party` being applied inconsistently. The filter log indicates that it is third party (and indeed it is a different domain), but the filtering seems act as if it's first party.

Indeed the behaviour is inverted on the iOS version. Is this a bug with the Mac version? Should I submit a report there?

PR #16998, opened 2018-05-26
Update specific.txt

mrtarlan:
Please remove these specific rules, and create ###koddostusplash general elemhide.


Alex-302:
less than even 10 sites. Not sure if this is required.

PR #17227, opened 2018-05-30
Updating for quora.com

aljelly:
(Not sure why the last line in the file is listed as changed, I didn't touch it)


Alex-302:
Hi
I need url of that page

aljelly:
For example, it's showing up here:
https://www.quora.com/How-can-I-report-Google-Ads-which-are-annoying

PR #19215, opened 2018-07-04
Improved matching of related domains

Grossdm:
On transip.* domains, e.g. transip.be or transip.eu, the following filter in cookies_general.txt prevents the site from accepting the user's acknowledgement of the cookie policy:
`/cookie-consent/*^$domain=~argeweb.nl`

There is an exception filter in cookies_whitelist.txt for just one of these domains:
`@@||transip.nl/services/cookie-consent/`

With the large number of related domains, I suggest a wildcard filter.


BlazDT:
It looks like they changed their site as that script is not requested anymore.

Grossdm:
@BlazDT, it looks like you are correct about the script.

Did you check to see whether the `/cookie-consent/*^$domain=~argeweb.nl` filter was removed? (I have not...)

BlazDT:
It is still in code as it blocks cookie scripts for other sites.
If the site is not using it anymore next Friday, I will remove initial exclusion rule from our filters and close this here.

Grossdm:
Thank you! This is my first time trying to make a contribution.
Baby steps... ☺

BlazDT:
Just seen it as a request, unblocked for every domain.

PR #21815, opened 2018-09-01
Attempted to add Tor domains to some entries

DandelionSprout:
As I have recently begun to experiment with .onion domains in Tor Browser (e.g. for the purposes of https://github.com/DandelionSprout/adfilt/blob/master/BrowseWebsitesWithoutLoggingIn.txt), I've learned that filter entries that apply to e.g. Facebook and NY Times, don't automatically apply to their .onion domains.

So I decided to attempt to add those domains to the relevant entries, and am suggesting that it could become some sort of policy to add .onion domains to theorethical future Facebook, NY Times and DuckDuckGo entries.


Alex-302:
added

PR #24931, opened 2018-11-06
Removed duplicates

bogachenkove:


krystian3w:
<i>

You can still compress `0px` most of the css filters (not in attribute eg `[style]`).

`marign-right: 0px!important` to `marign-right: 0!important`
`padding-right: 0px!important` to `padding-right: 0!important`

</i>

Alex-302:
> You can

done

PR #25035, opened 2018-11-08
Update specific.txt

ghost:
These sites are no longer available or expired.Please remove.


Alex-302:
Done.
@PentagramTr better give us the dead domains list, instead of the rules.

PR #26112, opened 2018-11-27
taiphanmempc.net #23055

slavaleleka:
.


BlazDT:
Popup not showing, only in code. Scrolling still works after hiding it?

slavaleleka:
Yea, scrolling works just fine

Alex-302 (changes requested):

BlazDT (changes requested):
No need for "third-party" option if you already use $domain.
Also please use ^$domain=   instead of only $domain

Alex-302 (approved):

PR #26206, opened 2018-11-29
ikea.com.tr #24050 +

slavaleleka:
also added cookie banner blocking rule for ikea.com (exactly for ikea.com/ru/ru)


BlazDT (approved):
Ikea.com not reproduced

Alex-302 (approved):

Issue #24494, opened 2018-10-27
pornxs.com [NSFW] [Ad String]


Nzzyme:
**Site**: https://pornxs.com/video.php?id=3258198&utm_source=awn-rse&utm_medium=inc&utm_campaign=20171124

**Issue**: Please remove the ad text on the page

**Screenshot**: http://prntscr.com/lb1p1s

**OS**:
> `Windows 7 (64 bit)`

**Browser**: 
>` Chrome 70.0.3538.77 (Official Build) (64-bit)`

**AG Version**:
> `AG for Chrome Browser Extension 2.10.13 `

**Filters**:
> `Base Filter, Spyware, Social Media, Annoyances, Mobile Ads, Adblock Warning Removal List, Malware Domains, Spam 404
[No user filter used | Phishing & Malware Protection Enabled] `

bogachenko:
check
`pornxs.com##.video__container > .video__inner > .video__content + :last-child`

Nzzyme:
Still seeing some other ad text strings

https://pornxs.com/21729115

http://prntscr.com/nje01i

PR #26454, opened 2018-12-03
pornxs.com #24494

slavaleleka:
.


BlazDT (changes requested):
Cookie rule already in Annoyances filter.

Alex-302 (approved):

PR #26590, opened 2018-12-06
japanese.engadget.com #24276

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/24276

remain of element under the header can not be hidden because of EnglishFilter/whitelist rule
`engadget.com#@#aside[role="banner"]`


JapaneseFilter/sections/specific.txt
```diff
@@ -15,6 +15,7 @@ kyoko-np.net##div[class^="col-"] > div.ad
 sankei.com###sponsored
 ganganonline.com##header#sqexHeader-black_rsp
 asahi.com##.AdMod
+japanese.engadget.com##p.partner_container
```
Alex-302: that exclusion in our Eng filter. The rule with comment, so you can check if it is still required.

Alex-302 (changes requested):

JapaneseFilter/sections/specific.txt
slavaleleka: yea, it's still required, but it's not a problem anymore

Alex-302 (approved):

JapaneseFilter/sections/specific.txt
```diff
@@ -15,6 +15,8 @@ kyoko-np.net##div[class^="col-"] > div.ad
...
+japanese.engadget.com##aside#ldrBoardAd
```
Alex-302: Please check if this rule required on engadget.com

slavaleleka: there are no such elements on engadget.com

Alex-302 (approved):

PR #26623, opened 2018-12-07
togetter.com #24279

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/24279

cannot reproduce the issue dated 23 oct


BlazDT (commented):
Also initial issue not reproduced, but please add this rule to hide ad left-over on desktop version on main page:

`togetter.com##.ad_topics_custom`

Alex-302 (approved):

Issue #27235, opened 2018-12-20
Tapatalk banner on mobile versions of forums


adranger:
Tapatalk ads are displayed to mobile Firefox users.

Blocking `/smartbanner/appbanner.`  fixes this.

Example pages:

    http://www.reefclub.net.br/community/index.php
    http://forum.brfoto.com.br/
    https://forum.lolesporte.com/viewforum.php?f=8
    http://forumchaves.com.br/

![tapatalk](https://user-images.githubusercontent.com/35470489/49576380-32be2700-f92c-11e8-9d41-25f134dbc706.png)

Firefox 63 for Android
uBlock Origin 1.17.4
Adguard Mobile Filters list

BlazDT:
Please enable Annoyances filter and update.

PR #28142, opened 2019-01-10
updated /mobiquo/smartbanner/appbanner.

ghost:
Removed "js" for this rule to block css files too. Examples: 

`http://forum.brfoto.com.br/`
`http://www.reefclub.net.br/community/index.php`


Alex-302:
@adranger in some cases CSS should not removed.

adranger:
@Alex-302 I don't think there's such a case because blocking JS prevents the Tapatalk banner to be displayed completely (https://github.com/AdguardTeam/AdguardFilters/issues/27235). The CSS files for that plugin are just unnecessarily loaded in the background, and I just wanted to save people some mobile bandwidth. Anyway, it's your choice.

PR #28330, opened 2019-01-14
faapy.com #25207

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/25207

### DIFF
--- a/EnglishFilter/sections/specific.txt
+++ b/EnglishFilter/sections/specific.txt
@@ -164,6 +164,8 @@ kissanime.ac###adlink2
 ||vcdn.rivertraffic.com/*.mp4$domain=viptube.com
 masseurporn.com##.video_cums
 masseurporn.com##.video_cums1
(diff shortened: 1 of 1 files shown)


EnglishFilter/sections/specific.txt
```diff
@@ -164,6 +164,8 @@ kissanime.ac###adlink2
 ||vcdn.rivertraffic.com/*.mp4$domain=viptube.com
 masseurporn.com##.video_cums
 masseurporn.com##.video_cums1
+faapy.com##a[href="https://faapy.com/cs/mylf/"]
+faapy.com##.container > .textlink
```
Alex-302: use more specific selector in this case. Something like this
`.footer > div.container > div.textlink`

Alex-302 (changes requested):

Alex-302 (approved):

PR #30207, opened 2019-02-27
Safari filter rework

Alex-302:
Copied to Base filter(content_blocker.txt)


ameshkov:
I just realized that we should prohibit optimization for all these rules in the `content_blocker.txt`

@Mizzick any idea how can we do it?

EnglishFilter/sections/content_blocker.txt
```diff
@@ -0,0 +1,1057 @@
+!
+! Section contains list of the rules for content blockers
+! It will be used by AdGuard for Android and AdGuard for iOS/Safari
+!
+! *** CAUTION: all rules are wrapped by AG directive. Don't remove it, and don't add another directives***
+!
+#if (ios || ext_safari || ext_android_cb)
```
AdamWr: I'm not sure, but shouldn't it be:
`#if (adguard_app_ios || adguard_ext_safari || adguard_ext_android_cb)`
instead of:
 `#if (ios || ext_safari || ext_android_cb)`?

Alex-302: sure. Fixed

BlazDT (approved):

BlazDT (approved):

EnglishFilter/sections/content_blocker.txt
AdamWr: Ok :)
One more thing, I think it should be `!#if` and the same for `!#endif`

Alex-302: done)

AdamWr (approved):

ameshkov:
Let's wait for @Mizzick to implement necessary changes to the filters compiler

ameshkov (approved):

PR #31913, opened 2019-04-01
fix stealth mode #31887 t-mobile.cz

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/31887

there no stealth mode parameters needed to be enabled


BlazDT (changes requested):
Disabling stealth for the whole site seems like overkill.

@@||t-mobile.cz/sms^$stealth

BlazDT (approved):

AdamWr (approved):

PR #32265, opened 2019-04-08
fix #29071 nymag.com

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/29071


AnnoyancesFilter/sections/popups.txt
```diff
@@ -96,6 +96,7 @@ atlanticleague.com##.incap_page-tooltip
 58pic.com##.activitySettings-fixedBottom
 global.llbean.com##.subscribe-modal
 global.llbean.com##.ui-widget-overlay
+nymag.com##.modal.right-aligned
```
AdamWr: As I understand this rule hides newsletter popup, if so, it looks okay, but maybe it would be better to use:
`nymag.com##.newsletter-slideout`

AdamWr (approved):

PR #32713, opened 2019-04-18
pandora.com.tr - Unlocking locked actions

BlazDT:
pandora.com.tr - Unlocking locked actions
https://github.com/AdguardTeam/AdguardFilters/issues/32710


AdamWr:
Maybe it would be better to check if element exists?
Otherwise on main page (https://www.pandora.com.tr/) will be error in console:
`Uncaught TypeError: Cannot set property 'ondragstart' of null`.

PR #34992, opened 2019-06-09
spiegel.de

KAMiKAZOW:
Blocks this:

Screenshot_20190609-171324_Firefox


KAMiKAZOW:
Not sure where that edeka.de line came from. Didn't touch that at all.

Alex-302:
fixed

PR #35089, opened 2019-06-11
fix ellegirl.ru #33353

zloyden:
https://github.com/AdguardTeam/AdguardFilters/issues/33353

### DIFF
--- a/SocialFilter/sections/specific.txt
+++ b/SocialFilter/sections/specific.txt
@@ -61,6 +61,7 @@ vgorode.ua##.social_plugins
 fakty.com.ua##.like_box > li:not(.f_right)
 afmen.online##.wrap-socials_webnavoz
 ipress.ua##.yashare
+ellegirl.ru##.social-container
 tsn.ua##.js-fb-message
 tsn.ua##.wtf_bc--feed-item--share
 sys-team-admin.ru##.likes-block_center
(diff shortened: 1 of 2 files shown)


AdamWr:
In my opinion rules looks okay, but if I'm not wrong, there are also ad leftovers on this website.

image

</details>

I think that this rule should fixes these ad leftovers:
`ellegirl.ru##.creative-wrapper`

Could you please check it too?

Alex-302 (changes requested):
please block ad leftovers

SocialFilter/sections/specific.txt
```diff
@@ -61,6 +61,7 @@ vgorode.ua##.social_plugins
 fakty.com.ua##.like_box > li:not(.f_right)
 afmen.online##.wrap-socials_webnavoz
 ipress.ua##.yashare
+ellegirl.ru##.social-container
```
Alex-302: just links to blogs. Usually we don't block them

Alex-302 (changes requested):

SocialFilter/sections/specific.txt
BlazDT: They are hiding the widgets which show e. g. their posts, not a link to their e. g. facebook site.

Alex-302: I see. @zloyden please also block `http://platform.twitter.com/widgets.js` (add domain to existing rule)

PR #38523, opened 2019-08-13
fix vinnytsianews.com https://forum.adguard.com/index.php?threads/htt…

zloyden:
https://forum.adguard.com/index.php?threads/https-vinnytsianews-com.34339/


RussianFilter/sections/specific.txt
```diff
@@ -19,6 +19,8 @@ Ly9hbi55YW5kZXgucnUv
 ||cdn.*/libs/b.js$third-party,domain=nnm-club.lib|nnm-club.me|nnm-club.tv|nnmclub.to|seasonvar.ru|mintmanga.com|realmadrid.ru
 !
 dnepr.info##.sidebar > .blue-bann
+vinnytsianews.com##a[href*="yanina.life/"]
```
Alex-302: you can use `custom_html-` and `text-`. Useful elements has non default IDs.

Alex-302 (changes requested):

Alex-302 (approved):

RussianFilter/sections/specific.txt
Alex-302: > Useful elements has non default IDs.

** not on all sites.

PR #38534, opened 2019-08-13
fix xk5.com.ua https://forum.adguard.com/index.php?threads/https-xk5-com-ua.34351

tvinzz:
https://forum.adguard.com/index.php?threads/https-xk5-com-ua.34351/


Alex-302 (changes requested):
I see missed ads. Please check again.

slavaleleka (approved):

PR #38897, opened 2019-08-20
fix #38837_auchan.ua

tvinzz:
https://github.com/AdguardTeam/AdguardFilters/issues/38837


BlazDT (changes requested):
Please also block browser notification about location.

Alex-302 (approved):

PR #40019, opened 2019-09-06
extra/#24 pogoda.yandex.ru

tvinzz:
https://github.com/AdguardTeam/AdGuardExtra/issues/24


ameshkov:
It'd be better to handle it in Extra

PR #42454, opened 2019-10-16
Too broad filter, blocking legitimate scripts

indeed-a-genius:


Alex-302:
Hi. We need few examples to check the rule.

indeed-a-genius:
Don't have any examples. Stumbled upon this issue when developing my own website. The rule is too broad regardless.

Alex-302:
I need at least one example)

indeed-a-genius:
on https://m.lh1backup.pw/city/Dubai
these scripts are blocked:
https://m.lh1backup.pw/client/dist/front/js/main.js?ex=15197
https://m.lh1backup.pw/client/dist/m/js/vendor.js?ex=e03d0
https://m.lh1backup.pw/client/dist/m/js/app.js?ex=e03d0

Alex-302:
I mean real example.

indeed-a-genius:
Please check your e-mail notification for the original urls, I edited them
after for privacy reasons.

krystian3w:
obraz

with this planed removed filter.

Alex-302:
@indeed-a-genius that rule breaks only your sites. As I see, you don't want to see exclusions in the filter, but I can't remove this rule because it blocks ads. Changed it, will be applied to third-party requests only.

PR #44734, opened 2019-11-25
oload.party whitelist specific 2 filter

Niush:
oload.party whitelist specific 2 filter for Base Filter to work properly

***Description***:
* **Current behaviour**: 

image
</details><br/>

* **Expected behaviour**: 

image

</details><br/>

***System configuration***

**Filters:**

Browser:                               | Chrome 78.0.3904.108
AdGuard version:                       | 3.3.4
Filters enabled:                       | AdGuard Base Filter


Alex-302:
Hi. Please provide sample  link.
Also, instead of exclusion better remove domain from original blocking rule.

Niush:
Removed filters instead of exclusion

Sample link [https://pixat.ml/](https://pixat.ml/) after playing a video, if Base Filter is active overlay is present which goes away if filter is removed. Allowing Play button to be clickable.

AdamWr:
I think that direct link will be better:
https://oload.party/loadsource.php?server=12&id=3593673&token=xHQlhemxrX8AZY2QFgdw9i7O7zauDYT3OO8Gery/tXeDIL5B2ROsMvNowNYz11mwsEixmT74vZIAc5wAyASxtettWt1Vh6r7VpMIar/nwBs0cgOxKFdKvO5YokdBbN2KRTejnTIQoMFaIU3TTVpmrrZ/nN+svNJJWuQM6di+rllyccJyz/6+nQXJg9g75zAou0HUcgue7iROUls7q92S1w==

I get a popup/popunder (on first visit) if these rules are disabled.

What about just hidding this overlay?
For example:
```
oload.party##body > div[style^="z-index: 999"][style*="position: absolute;"]
mixdrop.co##div[style^="position:absolute;"][style*="z-index:"][onclick]
```

Alex-302:
> What about just hidding this overlay?

as I remember(few years ago:) ), it can break player on iOS.

AdamWr:
As I recall it was on websites which was using `propellerads` script (`zfgloaded`), this script is different.
Anyway, if it's a problem on iOS, we can add hints, I think.
```
!+ NOT_PLATFORM(ios, ext_android_cb)
oload.party##body > div[style^="z-index: 999"][style*="position: absolute;"]
mixdrop.co##div[style^="position:absolute;"][style*="z-index:"][onclick]
```
or add:
`oload.party#%#//scriptlet("abort-current-inline-script", "decodeURI", "4zz.")`
seems to works fine too.

Alex-302:
Better check on iOS)

AdamWr:
It seems that at the moment this video doesn't work on iOS.
After adding these rules video works.
There are also popups with and without these rules, but I guess it can't be fixed on iOS.

Niush:
I don't really know much about advanced filters. I would appreciate any working solutions, if you can.

AdamWr:
@Niush  Could you please check if it work on your side if you add these rules to user filter?
```
oload.party##body > div[style^="z-index: 999"][style*="position: absolute;"]
mixdrop.co##div[style^="position:absolute;"][style*="z-index:"][onclick]
```
or instead of first rule, this:
`oload.party#%#//scriptlet("abort-current-inline-script", "decodeURI", "4zz.")`

Niush:
@AdamWr 

oload.party##body > div[style^="z-index: 999"][style*="position: absolute;"]
mixdrop.co##div[style^="position:absolute;"][style*="z-index:"][onclick]

this seems to be working...

Can you also explain what does the scriptlet thing does?

_Sent from my Desire 830 dual sim using [FastHub](https://play.google.com/store/apps/details?id=com.fastaccess.github)_

AdamWr:
This scriptlet should abort one of the inline script, which seems to be related to popups.

image

</details>

There is more about scriptlets - https://github.com/AdguardTeam/Scriptlets
https://github.com/AdguardTeam/Scriptlets#abort-current-inline-script

AdamWr:
OK, I removed this rule:
```
oload.party#%#(function(){var c=document.addEventListener;document.addEventListener=function(a,b,d,e){"mousedown"!=a&&-1==b.toString().indexOf('Z4P')&&c(a,b,d,e)}.bind(document); var b=window.setTimeout;window.setTimeout=function(a,c){if(!/\..4P\(|=setTimeout\(/.test(a.toString()))return b(a,c)};})(); 
```
and added these:
```
oload.party#%#//scriptlet("abort-current-inline-script", "decodeURI", "4zz.")
oload.party##body > div[style^="z-index: 999"][style*="position: absolute;"]
mixdrop.co##div[style^="position:absolute;"][style*="z-index:"][onclick]
```
it should be okay.
Update should be available in 1-2 hours.

Thank you for your reply.

PR #46177, opened 2019-12-18
fix #46150 womenshealthmag.com

### LINKED ISSUE #46150: womenshealthmag.com
### Issue URL (Annoyance)
[https://www.womenshealthmag.com/es/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.womenshealthmag.com%2Fes%2F)

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
Platform: | Ext
AdGuard version: | 3.3.4
Browser: | Chrome
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> AdGuard Annoyances<br/><br/><b>Language-specific:</b><br/> AdGuard Spanish/Portuguese

Sergey-Lyapin:
#46150

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -1630,7 +1630,7 @@ reddit.com###SHORTCUT_FOCUSABLE_DIV > div > div:not([class]):not([id]):not([styl
 forticlient.com,fortiguard.com,fortinet.com###__agreementBar
 3m.com.tr,command.com.tr,virginmedia.com,francetvinfo.fr,greenmangaming.com,logitech.com,logitechg.com,webex.com###__tealiumGDPRecModal
 thequietus.com###_evidon-background
-womenshealthmag.com,esquire.com,geek.com,ign.com,pcmag.com,speedtest.net###_evidon-barrier-wrapper
+esquire.com,geek.com,ign.com,pcmag.com,speedtest.net###_evidon-barrier-wrapper
 telegraph.co.uk,humblebundle.com,last.fm,linuxtoday.com,pcmag.com,techrepublic.com,zdnet.com###_evidon_banner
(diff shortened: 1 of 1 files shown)


AdamWr:
@Sergey-Lyapin Could you please check if with this rule, image(s) from Instagram loads fine here - https://www.womenshealthmag.com/es/moda-belleza/a28886769/365-dias-peinado-igual-recogido-salud-pelo-crecer/

It seems that it doesn't load until cookie is accepted, at least on my end.

image

</details>

So maybe it would be better to add something like this:
`womenshealthmag.com#%#document.cookie = "_evidon_consent_cookie=1; path=/;";`
There is already similar rule in Annoyances filter.

BlazDT:
Can confirm the non-loading picture of Instagram without accepting cookie.

zloyden:
> So maybe it would be better to add something like this:
> `womenshealthmag.com#%#document.cookie = "_evidon_consent_cookie=1; path=/;";`
> There is already similar rule in Annoyances filter.

I was able to reproduce it, so, I guess we must limit the blocking cookie rule with platform for content blockers.

Sergey-Lyapin:
@AdamWr @zloyden 
Could you take a look?

AdamWr:
I think that `womenshealthmag.com###_evidon-barrier-wrapper` is not needed if there is a JS rule.
By the way, AdGuard extension for Firefox and Opera can't update/download new JS rules, so if we need a cosmetic rule, then a hint should looks like that:
`!+ NOT_PLATFORM(ios, ext_android_cb, ext_opera, ext_ublock, ext_ff)`

AdamWr (approved):

zloyden (approved):

PR #47618, opened 2020-01-13
fix yandex popups

Sergey-Lyapin:


AnnoyancesFilter/sections/annoyances.txt
```diff
@@ -1172,6 +1172,12 @@ eda.yandex##.src-common-components-common-MobileAddressPromptBar--mobileAddressP
 !
 yadi.sk##.app-promo-banner
 !
+! Yandex.Tv teaser banner
```
Alex-302: what abour another regional domains?

Alex-302 (approved):

PR #48732, opened 2020-01-27
24smi.org [direct]

zloyden:
Direct appears after several refreshes.
https://github.com/AdguardTeam/AdGuardExtra/issues/48#issuecomment-578855240


RussianFilter/sections/antiadblock.txt
```diff
@@ -1486,6 +1486,7 @@ $cookie=/^los\$/,domain=auto.ru
 ! makes kinopoisk.ru redirect to sso.kinopoisk.ru and back endlessly
 ! ||kinopoisk.ru^$cookie=/consultant|introductory|knowledge|transfer|technology|workplace/
 !
+24smi.org#%#//scriptlet("set-constant", "Object.prototype.createAdvList", "undefined")
```
Alex-302: is the second rule still required?

zloyden: Yes, we need both of them.

PR #49406, opened 2020-02-07
fix businessman.ru #49254

Sergey-Lyapin:
#49254


RussianFilter/sections/specific.txt
```diff
@@ -20,6 +20,8 @@ mathprofi.ru##table[width][align="center"] > tbody > tr > td[valign="top"] > p[a
 kulturologia.ru##.post > div.content > p + center
 customgames.ru##.p-body-content > div.p-body-pageContent + center
 avtozam.com##.adhelper_inview
+businessmens.ru##.top-bnr
```
zloyden: First rule doesn't need, second rule should be moved to Annoyances.

zloyden (changes requested):

zloyden (approved):

PR #49442, opened 2020-02-08
deti.mail.ru [direct]

zloyden:
https://github.com/AdguardTeam/AdGuardExtra/issues/63
Scriptlet doesn't work on my end and I was able to reproduce it only in app.
<details><summary>with rule:</summary>

image
</details><br/>


zloyden:
According user's [answer](https://github.com/AdguardTeam/AdguardFilters/issues/48225#issuecomment-587068599) - ads don't reproduced. On my end it's the same.

PR #50143, opened 2020-02-19
Add obfuscated Appodeal backend

nyancrimew:


deletescape:
Use of this can be observed in this (adware) app for example: https://play.google.com/store/apps/details?id=com.wallpaper.wallpapers.fortnite

Alex-302:
Thanks. 
Changed rules order and added as commit.

PR #51428, opened 2020-03-13
fix #51373 cnet.com

### LINKED ISSUE #51373: cnet.com
### Issue URL (Annoyance)
 [https://www.cnet.com/products/dell-optiplex-fx160-atom-230-1-6-ghz-monitor-none-series/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.cnet.com%2Fproducts%2Fdell-optiplex-fx160-atom-230-1-6-ghz-monitor-none-series%2F)
### Comment
Username: @bazilisek
> po načtení stránky se vpravo dole objeví video

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
Platform: | Ext
AdGuard version: | 3.3.8
Browser: | Chrome
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Other:</b><br/> Filter unblocking search ads and self-promotion<br/><br/><b>Language-specific:</b><br/> AdGuard German, <br/>AdGuard Spanish/Portuguese, <br/>EasyList Czech and Slovak, <br/>Official Polish filters for AdBlock, uBlock Origin & AdGuard

Sergey-Lyapin:
#51373

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -264,8 +264,7 @@ radeberger.com##.cookieabfrage
 montiemare.de#$##cookieNotice { display: none !important; }
 montiemare.de#$#.modal-open { overflow: visible !important; }
 montiemare.de#$##cookieNotice + .modal-backdrop { display: none! important; }
-cnet.com,zdnet.com#$##onetrust-consent-sdk { display: none !important; }
-cnet.com,zdnet.com#$#button#ot-sdk-btn { display: none !important; }
+cnet.com,zdnet.com###ot-sdk-btn
 expedia.nl,expedia.ch,expedia.de,expedia.it,expedia.fr,expedia.at,expedia.dk,expedia.es,expedia.no,expedia.fi,expedia.co.uk,expedia.ie,expedia.se###mad-client-gdpr-consent-banner
 expedia.be###gdpr-consent-banner
 picdumps.com###pcdc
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -3558,6 +3558,7 @@ bmwfaq.org#?#.Notices[-ext-has='a[href="//www.bmwfaq.org/help/cookies"]']
 !
 ! JS rules
 !
+cnet.com#%#AG_onLoad(function() { setTimeout(function() { var el = document.getElementById("onetrust-accept-btn-handler"); if(el) { el.click(); } }, 2500); });
```
AdamWr: There is already similar rule:
https://github.com/AdguardTeam/AdguardFilters/blob/1e57cc7c97f989d625832917399851e2fc0b648a/AnnoyancesFilter/sections/cookies_specific.txt#L3605
so please remove this old rule or change a time after which this function is executed, if it's necessary.

AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -264,8 +264,8 @@ radeberger.com##.cookieabfrage
 montiemare.de#$##cookieNotice { display: none !important; }
 montiemare.de#$#.modal-open { overflow: visible !important; }
 montiemare.de#$##cookieNotice + .modal-backdrop { display: none! important; }
-zdnet.com#$##onetrust-consent-sdk { display: none !important; }
-zdnet.com#$#button#ot-sdk-btn { display: none !important; }
+cnet.com,zdnet.com#$##onetrust-consent-sdk { display: none !important; }
+cnet.com,zdnet.com#$#button#ot-sdk-btn { display: none !important; }
```
AdamWr: If I'm not wrong, these rules hide cookie notice, so I think that we should add them only for platforms which can update/download new JavaScript rules.
For example, extensions for Firefox and Opera can't update/download new JavaScript rules (due to Mozilla/Opera policy), so this cookie notice will be hidden and some of website functionality will be broken/not work.

AnnoyancesFilter/sections/cookies_specific.txt
AdamWr: Or we can exchange this rule to something like this:
```
cnet.com#%#AG_onLoad(function(){if(-1==document.cookie.indexOf("OptanonConsent")){var a=new MutationObserver(function(){var b=document.querySelector("button#onetrust-accept-btn-handler");b&&(b.click(),setTimeout(function(){a.disconnect()},1E3))});a.observe(document,{childList:!0,subtree:!0});setTimeout(function(){a.disconnect()},1E4)}});
```

zloyden: In my opinion this rule is ok, we should exchange it to yours.

Sergey-Lyapin: > Or we can exchange this rule to something like this:
> 
> ```
> cnet.com#%#AG_onLoad(function(){if(-1==document.cookie.indexOf("OptanonConsent")){var a=new MutationObserver(function(){var b=document.querySelector("button#onetrust-accept-btn-handler");b&&(b.click(),setTimeout(function(){a.disconnect()},1E3))});a.observe(document,{childList:!0,subtree:!0});setTimeout(function(){a.disconnect()},1E4)}});
> ```

Perfect rule. I'll add it, thanks.

AnnoyancesFilter/sections/cookies_specific.txt
Sergey-Lyapin: It looks like that we don't need these css rules anymore (because your new js rule hides cookie-popup instantly), so I think that we should leave only rule `cnet.com,zdnet.com###ot-sdk-btn` for hiding "manage cookies" button.

AdamWr: Ok, I think that it will be fine.

AdamWr (approved):

PR #52108, opened 2020-03-24
Remove bugsnag.com from tracking_servers.txt

keeganlow:
The rationale for this change is the same as that expressed in https://github.com/AdguardTeam/AdguardFilters/commit/c395ebc750b0385cec4a0342a35be94b35f2dcdf which removed crashlytics from this list.


Alex-302:
Done

PR #52626, opened 2020-03-31
Update specific.txt

ghost:
Obsolete rules.
I found these rules from Adguard Forum and checked again. Most rules were added in 2015 and 2016. Some sites have been closed, some have changed their designs. 

---

I just added "s" to the rule below.

`dogruses.com##a[href="https://canliradyodinle.gen.tr"]`  | `http` > `https`


AdamWr (approved):

AdamWr:
Thank you.
I re-added `dizikiyafet.com##a[href^="http://tr.rdrtr.com/aff_"]` and `internetoku.com##a[href^="http://www.internetoku.com/showad.asp"]`, because it seems that these rules still work.

PR #53873, opened 2020-04-17
Migrated many entries from "Adguard Annoyances Filter - Remains" to ad-specific lists.

DandelionSprout:
(This is an unusual kind of entry PR, so I'll disregard the repo's usual issue template this time.)

During while my various excursions deep into the thousands of lists out there, I've very slowly learned that AdGuard Filters have been placing ad placeholders in AdGuard Base **and** in AdGuard Annoyances Filter, with the distiction between them being far from obvious.

Eventually I remembered to ask about in the listmaker Slack group, and was told by Alex-302 that ad placeholders are to be treated the same way as ads, and that most of https://github.com/AdguardTeam/AdguardFilters/commits/master/AnnoyancesFilter/sections/remains.txt is just an ancient deprecated relic.

However, since a cleanup of that latter file did not seem to be high up on you guys' priorities, I decided to do a partial cleanup myself, to the degree it seemed feasible to do so without having to double-check every single one of the ~350 sites individually.

Edit: It has also been pointed out that the entries in the "Remains" file are usually really, really old and pointless. This PR assumes that the entire file won't simply be declared useless and deleted in one big sweep.

### DIFF
--- a/AnnoyancesFilter/sections/remains.txt
+++ b/AnnoyancesFilter/sections/remains.txt
@@ -3,51 +3,34 @@
 !
 voxc.org###link-view > p ~ hr
 voxc.org##.box-main > .banner ~ hr
-abhala.de##.hotinfo_top
 unlockapk.com##.StlyeSocial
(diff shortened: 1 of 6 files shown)


Alex-302:
We will check it.

chirag127:
.

PR #54522, opened 2020-04-28
Block social staff and tracking pixel at ad-ron.jp

Yuki2718:

***Description***:
* **Current behaviour**: 

URL `https://ad-ron.jp/?p=16196`

ad-ron

</details><br/>

* **Expected behaviour**: 

Social-share button and the leftover of social-bar should be blocked by Social filter, as well as tracking pixel by Tracking Protection filter.

Just visit the site with according filter lists.

***System configuration***

**Filters:**

Base, Tracking, Social, Annoyances, & Japanese

Browser:                               | Firefox 75.0
AdGuard version:                       | 3.4.19
Filters enabled:                       | Above

### DIFF
--- a/SocialFilter/sections/specific.txt
+++ b/SocialFilter/sections/specific.txt
@@ -1621,6 +1621,8 @@ ganganonline.com##.gn_detail_header_sns
 m.bafrahaber.com##.sosyal-alan
 bafrahaber.com##.detay-sosyal-alan
 meduza.io##.Toolbar-root > ul.Toolbar-list > li:nth-child(-n+5)
+||ad-ron.jp/wp-content/themes/adron/images/line-share-
+ad-ron.jp###share2
 newsweekjapan.jp###likeTool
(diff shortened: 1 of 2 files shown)


Yuki2718:
AFAIK `/line-share-` is not a common pattern so I though it's not suitable to general_url.txt, but forgive me if I made the commit in the wrong file.

Alex-302:
Added a bit  different rules.

Alex-302:
Also blocked widget with tracking(like taboola)

Also I see `js.dsp.reemo-ad.jp`. Looks like a tracker. Not sure.

Yuki2718:
> 
> 
> Also blocked widget with tracking(like taboola)
> 
> Also I see `js.dsp.reemo-ad.jp`. Looks like a tracker. Not sure.

I see `taxel.jp` on mobile, but somehow don't see `js.dsp.reemo-ad.jp`. Don't you see `tr.infopanel.jp`?

PR #54599, opened 2020-04-29
Fix text selection on as-selection.net: #54357

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/issues/54357#issuecomment-621240807


Alex-302 (approved):

Alex-302:
Bette use comment like that `Fix #54357 text selection on as-selection.net`
It will close existing issue when merged.

Yuki2718:
> 
> 
> Better use comment like that `Fix #54357 text selection on as-selection.net`
> It will close existing issue when merged(if it is required).

Okay, thank you for the advice.

PR #56413, opened 2020-05-29
fix #56261 chip.com.tr

### LINKED ISSUE #56261: chip.com.tr
 [https://www.chip.com.tr/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.chip.com.tr%2F)

### Screenshots
<details>

  Screenshot 1
</details>
<details>

  Screenshot 2
</details>

### System configuration

Information | value
--- | ---
Platform: | Ext
AdGuard version: | 3.4.31
Browser: | Chrome
Stealth mode options: | Hide your search queries, Send Do-Not-Track header, Remove X-Client-Data header from HTTP requests, Strip URLs from tracking parameters, Self-destructing third-party cookies (1440), Block WebRTC, Hide your Referrer from third-parties
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> AdGuard Annoyances<br/><br/><b>Security:</b><br/> NoCoin Filter List<br/><br/><b>Language-specific:</b><br/> AdGuard Turkish

Sergey-Lyapin:
#56261

### DIFF
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -90,6 +90,7 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !
 employment.en-japan.com##.guideSignup.js_stickyUnit
+chip.com.tr##div[class^="_MediaWorksFooter_"]
 poder360.com.br#$#.jquery-modal.blocker { display: none !important; }
 poder360.com.br#$#body { overflow: visible !important; }
 fossil.com###email-sign-up
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -90,7 +90,6 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !
 employment.en-japan.com##.guideSignup.js_stickyUnit
-chip.com.tr##div[class^="_MediaWorksFooter_"]
 poder360.com.br#$#.jquery-modal.blocker { display: none !important; }
 poder360.com.br#$#body { overflow: visible !important; }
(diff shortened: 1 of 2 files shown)


AdamWr:
I thought that this element is already blocked by this rule - `||amazonaws.com/Gravity/doganfooter.js` from Turkish filter. Isn't it?

Sergey-Lyapin:
> I thought that this element is already blocked by this rule - `||amazonaws.com/Gravity/doganfooter.js` from Turkish filter. Isn't it?

It is visible in Safari with extension and iOS, this script loads even if I add the rule in user filter, strange.
P.S. this rule exists in optimized filters.

AdamWr:
What about something like:
`||storagemw.s*.amazonaws.com/Gravity/doganfooter.js`

Sergey-Lyapin:
> What about something like:
> `||storagemw.s*.amazonaws.com/Gravity/doganfooter.js`

This rule works, thank you, I'll change the original rule. So what is the difference? For iOS&Safari we should add basic rules with subdomains?

AdamWr (approved):

AdamWr:
From what I found:
>`||` on Safari matches only one level of subdomains
`||example.org` matches `test.example.org`, but it does not match `another.test.example.org`

So in case like that, we should add rules with subdomains for iOS and Safari.

Sergey-Lyapin:
Thanks

PR #56870, opened 2020-06-06
zidongdianji app ads

ameshkov:
Application:
https://www.coolapk.com/apk/212036

Rules by @thb007


MobileFilter/sections/specific_app.txt
```diff
@@ -678,3 +678,6 @@ _*sec.mp4$mp4,domain=ivi.ru
 ||log.kajicam.com^
 ! Google Maps Promoted Pins
 ||mobilemaps-pa.googleapis.com/google.internal.mothership.maps.mobilemaps.ads.v1.MobileMapsAdsService/ListPromotedPinAds^
+! ads in this app: https://www.coolapk.com/apk/212036
+||dm.toutiao.com$app=com.zidongdianji
+||snssdk.com$app=com.zidongdianji
```
Alex-302, line 683: snssdk.com is already blocked by Mobile Filter
dm.toutiao.com - it is blocked in Chinese HOSTS lists. Probably we can do the same.

PR #57882, opened 2020-06-23
perimeterx.com cookie notice

SeriousHoax:

***Description***:
* **Current behaviour**: 

image
</details><br/>

* **Expected behaviour**: 

image
</details><br/>

***System configuration***

**Filters:**

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -3457,6 +3457,7 @@ obozrevatel.com##div[style^="z-index:999999999999999"]
 fasttorrent.ru,fast-torrent.ru#$?#html:has(> body) { padding-top: 0!important; }
 asos.com##header > ._2pMurAl._2gYXe0w
 spiceworks.com##header > div.sui-site-message-container
+perimeterx.com###gatsby-focus-wrapper > div > .open
 bose.de##header > div[class*="--cookie"]
 alternate.de#$##header[style^="margin-top:"] { margin-top: auto!important; }
 lawdepot.co.uk#$#html { margin-top: 0!important; }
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -3457,6 +3457,7 @@ obozrevatel.com##div[style^="z-index:999999999999999"]
 fasttorrent.ru,fast-torrent.ru#$?#html:has(> body) { padding-top: 0!important; }
 asos.com##header > ._2pMurAl._2gYXe0w
 spiceworks.com##header > div.sui-site-message-container
+perimeterx.com###gatsby-focus-wrapper > div > .open
```
slavaleleka, line 3460: it works. but the rule tells nothing about the subject of blocking. I mean that it's not obvious **what** will be blocked. we are trying not to add such rules

there is the link inside this element so extended-css rule would be better instead of this one. can you try to do that by yourself? or you need some help?

slavaleleka (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
SeriousHoax: Ow I see. I would appreciate if you help me.

slavaleleka: something like this would be fine
```
perimeterx.com#?##gatsby-focus-wrapper > div > aside:has(> p > a[href="/legal/privacy/"])
```

SeriousHoax: Ok thanks. Done.

slavaleleka (approved):

Issue #58587, opened 2020-07-04
Mass entry submission to AdGuard Japanese


DandelionSprout:
[//]: # (***You can delete or ignore strings starting with "[//]:" They will not be visible either way.)

Hopefully I can be forgiven for skipping most of the issue template, since this entry submission involves no less than 47 websites.

Essentially, it has been brought to my attention today that *AdGuard Japanese* can appear to be pretty small (~1,600 entries), and that [an up-and-coming Japanese enthusiast believed back in March](https://github.com/brave/adblock-lists/issues/355#issue-588587897) that *AdGuard Japanese* had an underrepresentation of mainstream Japanese websites.

So I've decided to do something about it, by going through №30 through 100 of https://domaintyper.com/top-websites/most-popular-websites-with-jp-domain, and submitting 70 entries from a total of 47 websites, to help establish AdGuard Japanese as the best JP-sites-only list once and for all. You are welcome to add whichever ones you want to add, but generally speaking, the more the better:

```
! ———Ads———
! https://zaif.jp/ (International site)
||cloudfront.net/images/banner^$image,domain=zaif.jp
zaif.jp##.banner-kv

! ———Sponsored article links———
! https://allabout.co.jp/
||adcontents.primead.jp^$image
allabout.co.jp##.aa_ad-home-thumb-small
allabout.co.jp##div[id^=pa_placement_]

! ———Arguable ad (Phone game listing-promo in a forum's baseball team section)———
! https://mixi.jp/view_community.pl?id=247
mixi.jp##.COMMUNITY_rightAreaBox:has(.COMMUNITY_rightGameAreaBox)

! ———Placeholders———
! https://www.daily.co.jp/
daily.co.jp##.pnl-aside

! https://www.homes.co.jp/
||adp.homes.co.jp^$image,domain=www.homes.co.jp
homes.co.jp##.adPremium

! https://www.2nn.jp/
2nn.jp##.ads

! https://www.jorudan.co.jp/
jorudan.co.jp###Ads_google_bottom_wide

! https://hatenablog.com/
hatenablog.com##.serviceTop-recommend-ad

! https://tver.jp/
||prd-static-cdn.tver.jp/web/TEAD_banner^$subdocument
###ADoverlay

! https://firestorage.jp/
firestorage.jp###mains > .minispace2

! https://weathernews.jp/onebox/ (International site)
weathernews.jp##.block:has(div[id^=div-gpt-ad])
[...]

PR #59205, opened 2020-07-13
Update #58587

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/issues/58587

Almost nothing, just found the tracking scripts are always third-party so thought adding `$third-party` will slightly increase performance. Also I found `||bidder.mediams.mb.softbank.jp^` in EasyPrivacy which I actually see at `https://rtrp.jp` and with Publicwww this: `http://104614.mediams.mb.softbank.jp/t/017/079/a1017079.js` too though not sure where it's used,  so made `js.mediams.mb.softbank.jp` more generic.


Alex-302 (approved):

Alex-302:
Ok. Thanks.

PR #60233, opened 2020-07-27
Went through most of the German entries in "Annoyances - Remains"

DandelionSprout:

***Description***:
* **Current behaviour**: 

I personally understand it as such that the "Remains" category in AdGuard Annoyances isn't really supposed to exist anymore, because such entries are nowadays meant be part of the regular anti-ad lists.

This time around I've actually browsed through the sites in question, one at a time, to see which entries I could reproduce (and move over to AdGuard German Filter), and which entries I couldn't reproduce (and thus were removed entirely).

I also stumbled across 6 new entries to add:
```
! https://www.autobild.de/artikel/audi-rs-4-vs.-rs-6-beschleunigung-0-100-v8-17086377.html
autobild.de##.rectangle
autobild.de##.partnerlabel
! https://www.derwesten.de/
||img.derwesten.de/img/incoming/*-315x80.jpg^
||img.derwesten.de/img/*-140x53.jpg^
||img.derwesten.de/img/incoming/*-315x80-$image
! https://www.tagesspiegel.de/
tagesspiegel.de###urban-teaser1
```

I suppose it deserves note that I had managed to forget that AdGuard had a German list, thus I only had EasyList Germany turned on during the testing... 😅 

* **Expected behaviour**: 

The entries in the "Remains" category are gradually re-reviewed and processed, until eventually all entries have been re-assigned and the "Remains" category can be deleted.

N/A

***System configuration***

**Filters:**

image

Operating system version (Android/iOS) | May 2020 Update
Browser:                               | Chrome x64
AdGuard version:                       | Nano Adblocker 1.0.0.152 + Nano Defender 15.0.0.201
Filters enabled:                       | See screenshot above
Filtering quality (Android only):      | N/A

### DIFF
--- a/AnnoyancesFilter/sections/remains.txt
+++ b/AnnoyancesFilter/sections/remains.txt
@@ -25,7 +25,6 @@ vividscreen.info##.page > .buddies
 vanar.io##.frame-data
 slain.io##.slain-adslot
 t-online.de##.Tvideosearch ~ #Tcontbox > #Tcontboxi > div[class]:not([id]):not([onfocus]):not(.Tinfinitlayer)
-t-online.de##div[data-dy-embedded-object]
 vivud.com##.closeContainer
 videam.party##.ads
 fmovies.is###movie > .widget > .widget > div[style*="height:166px;overflow:hidden"]
@@ -44,13 +43,10 @@ bigtitsxxxsex.com##.bnrba
 eurogamer.de##.mobile-mpu-wrapper
 t-online.de###T-Shopping
 memurlar.net##.NewsDetail > div[style*="width:580px; height:430px"]
-rtlnext.rtl.de##.rtli-master-morecontent__items > div.rtli-master-morecontent__teaser
-rtlnext.rtl.de#$#.rtli-master-section {margin-top: 0!important; }
 latimes.com##ark-ad-right
 latimes.com##ark-top-ad
 semprot.com##.semprotnenenmontok_adalah_pujaan_hatiku > div[style]:not([id]):not([class])
 semprot.com###semprotnenenmontok_f
-autobild.de###pagewrapper > #spacerElem
 aeroport-de-tunis-carthage.com##ins[id^="aswift_"]
 ashemaletube.com#$##site-wrapper {padding-top: 0!important; }
 ashemaletube.com#$##site {min-width: calc(100%)!important; max-width: calc(100%)!important;}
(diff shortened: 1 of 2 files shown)


Alex-302:
@BlazDT please take a look. As I see, some rules are not dangerous and hide visible ad leftovers.

AnnoyancesFilter/sections/remains.txt
```diff
@@ -73,7 +69,6 @@ bursa.com##.bg--gri[style="padding: 10px 0;"]
 m.pornxs.com###im_scroll_close_button
 m.smutty.com##.ui-content .center[style*="margin-"]:not([id])
 m.empflix.com##.bannerBlock
-t-online.de##div[data-dy^="Partner:"]:not([data-dy^="Partner:Telekom"])
```
BlazDT, line 76: Rule is found on t-online.de when I check the site, so please add it to German filter, thank you.

BlazDT (changes requested):

GermanFilter/sections/specific.txt
```diff
@@ -2123,7 +2123,7 @@ ruhr24.de,wetterauer-zeitung.de,giessener-allgemeine.de,fr.de,wlz-online.de,da-i
 wallstreet-online.de##body > #siteteaser
 wallstreet-online.de##.newsFlagSlider
 wallstreet-online.de###mooTickerContainer
-! https://github.com/AdguardTeam/AdguardFilters/pull/6023*
+! https://github.com/AdguardTeam/AdguardFilters/pull/60233
```
BlazDT, line 2126: No need to include link to pull request.

AnnoyancesFilter/sections/remains.txt
DandelionSprout: I do have the "Allow edits by maintainers" button turned on, you know. But sure thing, I'll take care of it. Give me 2min.

DandelionSprout: Oh, wait, now I remember. `t-online.de##div[data-dy*="Partner:"]` is already in both AdGuard German and in EasyList Germany, thus making the entry superfluous after all.

GermanFilter/sections/specific.txt
DandelionSprout: Has been removed now.

BlazDT (approved):

PR #61647, opened 2020-08-16
Fix in-site search in 700+ websites caused by marsflag.com

Yuki2718:

***Description***:
* **Current behaviour**: 

See this comment at EasyList: https://github.com/easylist/easylist/issues/5845#issuecomment-670581242. Blocking `marsflag.com`, or in case of AdGuard `c.marsflag.com` (no difference as far as this issue is concerned), breaks in-site search of virtually any websites I could find using Publicwww (700+ in number). I take `https://rent.toyota.co.jp/` as an example.

toyota

</details><br/>

* **Expected behaviour**: 

In-site search works.

toyota1

</details><br/>

Visit the page, copy and paste `カローラ` into the search box marked in the first SS, and press enter.

***System configuration***

**Filters:**

Base, Tracking, and Japanese

Browser:                               | Brave 1.12.112
AdGuard version:                       | 3.5.12

***Note***

`||marsflag.com/mf2file/site/ext/tr.js` is the real tracker which can safely be blocked so I added this instead.

### DIFF
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -22,6 +22,7 @@
 ||nd.demdex.net/event?d_mid=$empty,important,domain=9now.com.au
 ||adc.nine.com.au/?token=$empty,important,domain=9now.com.au
 !
+||marsflag.com/mf2file/site/ext/tr.js
 ||retailrocket.ru/content/javascript/tracking.js
 ||api.idg.zone/intextlinks-tracking/
 ||prometeo-media-service.com/assets/pixel.gif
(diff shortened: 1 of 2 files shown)


Alex-302:
Changed.

Yuki2718:
Somehow I don't see `tr.marsflag.com`. Thanks anyway.

Alex-302:
It is here
https://github.com/AdguardTeam/AdguardFilters/blob/533c639a2c50ca894a34a32a2eeb3d743e8e78ed/SpywareFilter/sections/tracking_servers_firstparty.txt#L226

Yuki2718:
> It is here

I meant I don't see any call to `tr.marsflag.com`.

toyota

Anyway it should be no harm, I believe you confirmed it doesn't break the site's functionality.

PR #62403, opened 2020-08-26
Update placeholder rules for itest.5ch.net

Yuki2718:
Placeholders, some are only seen when you searched something.

<details>

5ch-1

5ch-2

5ch-3

5ch-4

5ch-5

5ch-6

</details>

Hiding `##.js-cardview_ad-320x180` causes a layout problem, so limited to its child iframe. Given such a problem, I prefer not to consolidate them to fewer rules using `^=` or `*=`.


Alex-302 (approved):

Alex-302:
Thanks

PR #62425, opened 2020-08-26
'.js$empty' replaced by '.js$script,redirect=noopjs'

Alex-302:


BlazDT (approved):

slavaleleka (changes requested):
well done! 
 
I guess rules like `.../adsbygoogle.js$domain=nv.ua,empty` should be replaced as well. check via regexp search `\.js\$(.+)empty`

Alex-302:
> I guess rules like `.../adsbygoogle.js$domain=nv.ua,empty` should be replaced as well. check via regexp search `\.js\$(.+)empty`

I know, but too many changes. Will do it by parts.

zloyden (approved):

PR #64154, opened 2020-09-18
Update #58587 (fix uwl.weblio.jp layout)

Yuki2718:
It turned out the rule I suggested in the issue now breaks layout of `https://uwl.weblio.jp/vocab-index;`. This commit fixes it.

<details>

weblio

</details>


Alex-302 (approved):

Alex-302:
thanks

PR #64962, opened 2020-10-03
fix https://github.com/NanoMeow/QuickReports/issues/4778

Yuki2718:
https://github.com/NanoMeow/QuickReports/issues/4778
Fixed in Base as there is already an entry for its sister site (`9tsu.net`). `ad-vortex.com` is only on 9tsu.


Yuki2718:
@ryanbr I wonder if
```
||ad-vortex.com^
##.ads-system
```
can be added to EL.

EnglishFilter/sections/adservers.txt
```diff
@@ -6,6 +6,7 @@
 !! https://a.bestcontentitem.top/warp/*?r=
 !
 ||poemsdeucewee.com^
+||ad-vortex.com^
```
zloyden, line 9: We already have this rule in Easylist.

EnglishFilter/sections/specific.txt
```diff
@@ -855,7 +855,8 @@ kotaku.com##.js_movable_ad_slot
 forbes.com##fbs-ad + p > br
 swordz.io##div[style="width:300; height: 250; background-color: #000"]
 disasterscans.com##div[id^="media_image-"].no-icon.heading-style-1
-9tsu.net##.ads-system
+9tsu.net,miomio.us##.ads-system
+9tsu.net,miomio.us##.cactus-sidebar-content > .widget_text
```
zloyden, line 859: I don't see any other languages except Japanese. So, please move it to the AdGuard Japanese filter -> specific section.

zloyden (changes requested):

EnglishFilter/sections/adservers.txt
Yuki2718: Then that must be added after my commit.

EnglishFilter/sections/specific.txt
Yuki2718: Okay.

zloyden:
@Yuki2718 if you can't make changes, please tell me. I can do it directly.

Yuki2718:
> @Yuki2718 if you can't make changes, please tell me. I can do it directly.

Please do so, that'll be faster and mine is likely to cause conflict.

zloyden:
Added it to our filters, thank you!

PR #65583, opened 2020-10-14
Update #64737

Yuki2718:
Test URL is - or "was" - [NSFW] `https://jav.guru/88725/stars-208-i-love-my-teacher-and-i-hate-her-too-and-i-had-the-dqn-bad-boys-fuck-her-brains-out-yuna-ogura/` and need(ed) to choose UP server. This morning I saw requests such as
`https://morningsophisticatedcenter.com/56/39/b0/5639b0e88eb809d5d041d6d759e2dff5.js`
blocked by EL but indicating the regex needs to be updated. Unfortunately now I don't see it and instead `enormouslyauditorium.com`. I didn't take SS. I suggest to make this last case to set upper limit of characters, if we find 27+ cases probably it'll be better to remove the limit.

BTW default Stealth mode seems to break the UP server video.

PR #65841, opened 2020-10-18
Update #65583 regex for Popads (shorten.sh)

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/pull/65583
TBH I'm surprised to find a case requiring update so soon. On `https://shorten.sh/jjBarx` I see this:

shorten sh_jjBarx

29 char! Thankfully caught by a different but domain-specific regex. However, as I said earlier I suggest to remove the upper-limit.


Alex-302 (approved):

Alex-302:
Ok, let's try it)

PR #66047, opened 2020-10-21
fix #66023 excite.co.jp

### LINKED ISSUE #66023: excite.co.jp
### Issue URL (Ads)
 [https://www.excite.co.jp/news/article/Macotakara_40450/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.excite.co.jp%2Fnews%2Farticle%2FMacotakara_40450%2F)

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
AdGuard product: | AdGuard Browser extension v3.5.12
Browser: | EdgeChromium
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> AdGuard Annoyances<br/><br/><b>Language-specific:</b><br/> AdGuard Japanese

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/issues/66023

### DIFF
--- a/JapaneseFilter/sections/specific.txt
+++ b/JapaneseFilter/sections/specific.txt
@@ -3,6 +3,8 @@
 !
 ! /banner/banner_ - if it is necessary, make this rule domain specific
 !
+||excite.co.jp/jp/ox/
+www.excite.co.jp##div[id^=gpt_pc_news_]
 ||chaos2ch.com/rss.html
 ||player.daznservices.com^$domain=chunichi.co.jp
 cyzo.com##.rectangle
@@ -57,7 +59,6 @@ vippers.jp###relevantPost
 vippers.jp###topRss
 kitizawa.com###headerRSS
 kitizawa.com##div[style^="width:160px;height:600px;"]
-||image.excite.co.jp/jp/ox/gpt/display/pc/blog/pc_blog_billboard.js
 javtheater.com##.__isboostOverContent
 9tsu.net,miomio.us##.ads-system
 9tsu.net,miomio.us##.cactus-sidebar-content > .widget_text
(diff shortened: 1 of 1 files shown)


JapaneseFilter/sections/specific.txt
```diff
@@ -3,6 +3,8 @@
 !
 ! /banner/banner_ - if it is necessary, make this rule domain specific
 !
+||excite.co.jp/jp/ox/
+www.excite.co.jp##div[id^=gpt_pc_news_]
```
Alex-302, line 7: missed `"` in CSS selector

Alex-302 (changes requested):

JapaneseFilter/sections/specific.txt
Sergey-Lyapin: why with `www`?

Yuki2718: @Alex-302 Right, however, I found actually the cosmetic rule is not needed if /ox/ is blocked. I'll remove it entirely.
@Sergey-Lyapin Since excite has many subdomains, but no problem to remove www if it's better as the rule won't cause FP. Anyway I'll remove the rule in this case.

Yuki2718: Or do you prefer to keep the rule anyway even if it doesn't actually hide anything? The rule still hit, just not needed. If so, I'll simply add " and remove www.

PR #66342, opened 2020-10-25
Some fixes for JP filter cleanups

Yuki2718:
When I done them, I had Base, Spyware, and Japanese enabled but now noticed `business.nikkei.com##.cbCxense` is placeholder only if Spyware is enabled or `cxense.com` is blocked.

<details>

bnikkei

</details>

So I quickly checked the former cleanups too and found this is also the case to `koneta.nifty.com###custom_html-2` (placeholder only if `rtoaster.jp` is blocked). So moved them to Spyware. Also in the check process I happened to come across potential FP by `ascii.jp##.lnkBanner` (link banners to websites of the same corporate group)

<details>

ascii

</details>

So removed this too.


Alex-302 (approved):

Alex-302:
Thanks.

PR #67549, opened 2020-11-13
fix https://nandetribune.com/ anti-adblock

Yuki2718:

***Description***:
* **Current behaviour**: 

URL: `https://nandetribune.com/`

nandetribune

</details><br/>

Visit the site.

***System configuration***

**Filters:**

Base, Tracking, and Japanese

Browser:                               | Brave 1.16.72
AdGuard version:                       | 3.5.20


Alex-302:
Fixed.

Yuki2718:
generichide is not enough for content blocker. I followed past fix https://github.com/AdguardTeam/AdguardFilters/issues/33968 but sure scriptlet is an option for other platforms.

PR #67786, opened 2020-11-16
rakuten.com

ghost:
### URL: `https://www.rakuten.com`

Before:
before
After:
After

### Log:
Logs


Yuki2718:
See https://github.com/uBlockOrigin/uAssets/issues/5764
When it comes to trackers in shopping sites, checking appearance is not enough - we need to check operations as far as possible.

piquark6046:
> 
> 
> See [uBlockOrigin/uAssets#5764](https://github.com/uBlockOrigin/uAssets/issues/5764)
> When it comes to trackers in shopping sites, checking appearance is not enough - we need to check operations as far as possible (with and without login).

I am sorry for what I did not check that issue.
And, thank you for your comment.

PR #68631, opened 2020-11-27
Remove stockmann.com cookie dialog and allow scrolling

ollitapa:
***Description***:
Remove stockmann.com cookie dialog and allow scrolling

* **Current behaviour**: 

Cookie dialog shows and not able to scroll.

* **Expected behaviour**: 

No dialog.

* **Fix**:

Filter out the cookie div and allow scrolling on body. They use "position": fixed; to disable scrolling and take it away when dialog is accepted.

***System configuration***

**Filters:**

AdGuard Annoyances filter
Fanboy's Annoyances

Operating system:                      | iOS, Mac
Browser:                               | Safari
AdGuard version:                       | 1.8.6
Filters enabled:                       | yes

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -5056,8 +5056,8 @@ vi.nl#%#AG_onLoad(function() { if(window.location.href.indexOf("cookies") != -1)
 volkskrant.nl#%#AG_onLoad(function() { setTimeout(function() {if(document.getElementsByClassName("modal-overlay")[0].isvisible = true){document.getElementsByClassName("button button--accept")[0].click();}}, 300); });
 ziekenhuis.nl#%#AG_onLoad(function() { setTimeout(function() {var check = document.getElementsByClassName('cookiewall')[0]; if (check){document.cookie = "acceptcookies=true"; location.reload();}; }, 300); });
 !
-stockmann.com#$#body { position: static; }
-stockmann.com##div.cookie-dialog-screen.open
+stockmann.com#$#body { position: static!important; }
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -5056,6 +5056,9 @@ vi.nl#%#AG_onLoad(function() { if(window.location.href.indexOf("cookies") != -1)
 volkskrant.nl#%#AG_onLoad(function() { setTimeout(function() {if(document.getElementsByClassName("modal-overlay")[0].isvisible = true){document.getElementsByClassName("button button--accept")[0].click();}}, 300); });
 ziekenhuis.nl#%#AG_onLoad(function() { setTimeout(function() {var check = document.getElementsByClassName('cookiewall')[0]; if (check){document.cookie = "acceptcookies=true"; location.reload();}; }, 300); });
 !
+stockmann.com#$#body { position: static; }
+stockmann.com##div.cookie-dialog-screen.open
+!
```
slavaleleka, line 5061: for such cases rules should be type alike. I mean that both should be `#$#` and have `important` modifier. so eventually they should be:
```
stockmann.com#$#body { position: static!important; }
stockmann.com#$#.cookie-dialog-screen { display: none!important; }
```

could you please fix that?

slavaleleka (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
ollitapa: Oh, sure 👍

Alex-302 (approved):

Alex-302:
Thanks.

PR #69210, opened 2020-12-03
Fix #69089 ithome.com

### LINKED ISSUE #69089: ithome.com
 [https://www.ithome.com/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.ithome.com%2F)
### Comment
Username: @Zeuhx
> 
> --
> L'information ci-dessous est générée automatiquement, prière de ne pas la supprimer.
> 
> Referrer:
> 
> 
> Custom userscripts:
> https://userscripts.adtidy.org/release/assistant/4.3/assistant.user.js
> https://userscripts.adtidy.org/release/popup-blocker/2.5/popupblocker.user.js
> https://userscripts.adtidy.org/release/adguard-wot/1.0/wot.user.js
> 

### Screenshots
<details>

  Screenshot 1
</details>
<details>

  Screenshot 2
</details>

### System configuration

Information | value
--- | ---
Browser: | Edge 87.0.664.52 (Chromium-based)
AdGuard driver: | WFP
Stealth mode options: | Hide your search queries, <br/>Send Do-Not-Track header, <br/>Remove X-Client-Data header from HTTP requests, <br/>Strip URLs from tracking parameters, <br/>Self-destructing third-party cookies (180), <br/>Block WebRTC, <br/>Block Push API, <br/>Block Flash
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> Fanboy's Annoyances, <br/>Web Annoyances Ultralist, <br/>I don't care about cookies, <br/>EasyList Cookie List, <br/>AdGuard Annoyances<br/><br/><b>Other:</b><br/> Filter unblocking search ads and self-promotion, <br/>AdGuard DNS<br/><br/><b>Language-specific:</b><br/> AdGuard French, <br/>AdGuard Chinese

AdamWr:
Issue - https://github.com/AdguardTeam/AdguardFilters/issues/69089
Code - https://gist.github.com/AdamWr/f9e43774da32b1e6389c122fd8a73aa5#file-ithome_fix_empty_spaces-js

This problem occurs only if right column (one of the `#nnews > .t-b > ul.nl` elements) has less rows/children than the left column.
It seems that it depends on layout and it doesn't happen every time, sometimes is necessary to change a page (buttons with number at the bottom of the screenhot), sometimes it doesn't occur at all.
I tried to fix it with CSS rule, but it didn't work in all cases, maybe you will have a better idea.

image

</details>

### DIFF
--- a/EnglishFilter/sections/foreign.txt
+++ b/EnglishFilter/sections/foreign.txt
@@ -1000,8 +1000,9 @@ inmywordz.com##.acca-block
 manhuaren.com###lb-win
 ! https://github.com/AdguardTeam/AdguardFilters/issues/69089
 ! https://github.com/AdguardTeam/AdguardFilters/issues/62199
-ithome.com#?#.t-b.sel > ul.nl > li:has(> a[href*="/lapin."])
-ithome.com#$?##nnews > .t-b > ul.nl:nth-child(even):has(> li[style="display: none !important;"] ~ li[style="display: none !important;"]) { margin-bottom: 35px !important; }
+! https://gist.github.com/AdamWr/f9e43774da32b1e6389c122fd8a73aa5#file-ithome_fix_empty_spaces-js
+ithome.com#$?#.t-b.sel > ul.nl > li:has(> a[href*="/lapin."]) { remove: true; }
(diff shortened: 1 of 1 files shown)


ameshkov (changes requested):
1. Isn't it time to add these rules to the Chinese filter and not to Base?
2. This is a quite complicated rule for such a minor issue. Is it really justified to add it? Will we be able to maintain it?

Here's what I suggest: suggest the user add this rule to the user filter. If in the future there will be many similar complaints on the same empty placeholders, we may reconsider.

AdamWr:
Okay, so I will suggest this rule to the user and move extended CSS rule to Chinese filter (I guess it can be done without pull request).

PR #69235, opened 2020-12-04
Update #53079 coron.tech

Yuki2718:
The current solution works but no need to allow ad scripts on platforms supporting scriptlet. Plus some cosmetic staff. Of note, somehow now I don't see antiadb on `https://coron.tech/` even without solution but anti-adb scripts is still there so added scriptlet worked months ago (and should still work) just in case. `http://connect.coron.tech/` has a working anti-adb tho.


AdamWr (approved):

Sergey-Lyapin:
@AdamWr 
Hi) Do you see antiadblock on this site? I can't reproduce

AdamWr:
Hi:)
Yes, I do. On `http://connect.coron.tech/`

Yuki2718:
@Sergey-Lyapin As noted, no anti-adb on `https://coron.tech/` but I got few months ago and the script is still sitting there.

Sergey-Lyapin (approved):

Yuki2718:
Thx!

PR #69504, opened 2020-12-07
Fix for ilovepdf.com

rakleed:
- Remove the share buttons on the page with the converted file.


SocialFilter/sections/specific.txt
```diff
@@ -36,6 +36,7 @@ sonsaat.com.tr,okudunmuhaberi.com,spooor.com,mavididim.com.tr,filmfiz.net,filmw.
...
 !
 focus.ua##.c-post__social
+ilovepdf.com##.card
```
zloyden, line 39: `.card` isn't as unique as `.download__add`, so it's better to replace it to `ilovepdf.com##.download__add`.

zloyden (changes requested):

zloyden (approved):

SocialFilter/sections/specific.txt
zloyden: Nvm, I made changes.

Rakleed: @zloyden I checked again, and for me the `.download__add` rule does not work, but the `.card` rule works.

zloyden: It works fine on my end. Can you please check `ilovepdf.com#?#.container > div[class]:has(> div.card div.share)` in the User rules?

Rakleed: The rule `ilovepdf.com#?#.container > div[class]:has(> div.card div.share)` works for me.

zloyden: Fine, it was added recently. Filters update should be available in 1-2 hour.

Rakleed: Why does the `ilovepdf.com##.download__add` rule work only in the extension, but not in the application? Is this an app bug or platform limitation?

zloyden: The extension and the app have a different approach to blocking elements in such cases, but we have a possibilites to block it via app. Thanks for your feedback:)

PR #69641, opened 2020-12-09
fix ananweb.jp Brave anti-adb

Yuki2718:

***Description***:
* **Current behaviour**: 

`https://ananweb.jp/`

Brave anti-adb

ananweb

</details><br/>

Visit the site with any browser other than Brave and wait a second.

***System configuration***

**Filters:**

Base, TPL, & JP

Operating system:                      | Win 10
Browser:                               | Firefox 83.0
AdGuard version:                       | 3.5.23


Alex-302 (approved):

Yuki2718:
@Alex-302 Did you get notice about Akamai BM issue on EL repo?

PR #69671, opened 2020-12-10
fix #69656_mangaplus.shueisha.co.jp

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/issues/69656
Tried to avoid extended syntax for cookie banner, but there may be room for argument.

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -85,8 +85,7 @@ motorcyclenews.com,finanzen.net,vorname.com,jezebel.com,theonion.com,theinventor
 !
 !
 ||mangaplus.shueisha.co.jp/img/img_gdpr_
-mangaplus.shueisha.co.jp##.styles-module_message_WagEa
-mangaplus.shueisha.co.jp##.styles-module_message_WagEa + button[class^="styles-module_btn"]
+mangaplus.shueisha.co.jp##img[src^="/img/img_gdpr_"] + div[class^="styles-module_body_"]
 mangaplus.shueisha.co.jp##div[class^="styles-module_container_"] > div[class^="styles-module_yellow_"]
 bbc.com##div[class*="DialogBackdrop"]
 binance.com##.chat-widget + div.css-42befo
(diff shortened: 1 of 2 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -84,6 +84,9 @@ motorcyclenews.com,finanzen.net,vorname.com,jezebel.com,theonion.com,theinventor
 /wrapperMessagingWithoutDetection.*.js$script,domain=netzwelt.de
 !
 !
+||mangaplus.shueisha.co.jp/img/img_gdpr_
+mangaplus.shueisha.co.jp##img[src^="/img/img_gdpr_"] + div[class^="styles-module_body_"]
```
zloyden, line 88: I guess it's ok. Also you can try `mangaplus.shueisha.co.jp###app > div.vue-notification-group + div > div[class^="styles-module_container_"]:first-child` instead of them. Please check it on your end.

AnnoyancesFilter/sections/popups.txt
```diff
@@ -99,6 +99,8 @@ sodapdf.com##.be-tooltip
 !
 ! Specific sites
 !
+mangaplus.shueisha.co.jp#$#div[class^="index-module_container_"] { display: none !important;}
+mangaplus.shueisha.co.jp#$#body[style="position: fixed;"] { position: static !important; }
```
zloyden, line 103: Unfortunately, I don't see the popup.

AnnoyancesFilter/sections/cookies_specific.txt
Yuki2718: Actually I also thought about that, but `first-child` may be vulnerable to furure change of the site.

AnnoyancesFilter/sections/popups.txt
Yuki2718: I used Boston IP - though the site is co.jp inaccesible from Japane.

Yuki2718: I'll leave for a while, please freely close PR and fix on your end if appropriate.

zloyden: It's finally reproduced with US IP. The rules seem to be normal.

AnnoyancesFilter/sections/cookies_specific.txt
zloyden: Tbh, it's fair for every rule, but ok.

zloyden (approved):

PR #70208, opened 2020-12-17
Update specific.txt

gork7777:
I've added a few sites that are social widgets


Yuki2718:
I don't see `vatanbilgisayar.com##ul.social-networks.clearfix` unblocked. Even if unblocked, don't think ul and clearfix are necessary, better to add the domain to `radiohamburg.de##.social-networks`.
`covid19.saglik.gov.tr##ul.social-network` is out of scope for Social Media filter, they're the only follow buttons seen in the site. We block follow buttons only if they're redundant.

Yuki2718:
@Alex-302 At least `vatanbilgisayar.com##h4.footer-vatan-info__title.editable` is worth blocked, IDK why OP closed PR. But you may want to specify the filter bit more.

PR #70219, opened 2020-12-17
fix #70031 babesinporn.com

stanislav-atr:
#70031

PR #70308, opened 2020-12-18
Update https://github.com/AdguardTeam/AdguardFilters/issues/70031

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/pull/70219#discussion_r545828733
Please wait before merging, I'll check script too. In some cases visiting HP is not enough, only shown when redirected from sister sites.

### DIFF
--- a/EnglishFilter/sections/general_extensions.txt
+++ b/EnglishFilter/sections/general_extensions.txt
@@ -1734,7 +1734,7 @@ upload.ac#%#AG_onLoad(function() { setTimeout(function() { var el = document.get
 ! https://forum.adguard.com/index.php?threads/userscloud-com.32800
 userscloud.com#%#AG_onLoad(function() { setTimeout(function() { var el = document.getElementById("btn_download"); if(el) {el.removeAttribute("onclick");} }, 300); });   
 ! dancers `Popping Tool`
-jumboporn.xyz,freeuseporn.com,zazzybabes.com,vipergirls.to,imgixxx.com#%#//scriptlet("abort-current-inline-script", "loadTool", "/'popping'/")
+8boobs.com,babesinporn.com,bustybloom.com,hotstunners.com,rabbitsfun.com,silkengirl.com,jumboporn.xyz,zazzybabes.com#%#//scriptlet("abort-on-property-read", "loadTool")
 picbaron.com,porno-island.games,wantedbabes.com,consoletarget.com,celeb.today,ftopx.com,bunnylust.com#%#AG_abortInlineScript(/'popping'/, 'loadTool');
 ! healthinsider.online - popups
 healthinsider.online#%#window.open = function() {};
--- a/EnglishFilter/sections/general_extensions.txt
+++ b/EnglishFilter/sections/general_extensions.txt
@@ -1734,7 +1734,7 @@ upload.ac#%#AG_onLoad(function() { setTimeout(function() { var el = document.get
 ! https://forum.adguard.com/index.php?threads/userscloud-com.32800
 userscloud.com#%#AG_onLoad(function() { setTimeout(function() { var el = document.getElementById("btn_download"); if(el) {el.removeAttribute("onclick");} }, 300); });   
 ! dancers `Popping Tool`
-8boobs.com,babesinporn.com,bustybloom.com,hotstunners.com,rabbitsfun.com,silkengirl.com,jumboporn.xyz,zazzybabes.com#%#//scriptlet("abort-on-property-read", "loadTool")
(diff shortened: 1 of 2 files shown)


zloyden:
Ok, when your PR is ready to review, just removed `[WIP]` from the title.

Yuki2718:
I didn't see any `/strp_v` script.

zloyden:
> I didn't see any `/strp_v` script.

Do you mean in new domains or generally?

Yuki2718:
> > I didn't see any `/strp_v` script.
> 
> Do you mean in new domains or generally?

Generally, unless I missed.

zloyden:
Hm, for example

image

</details><br/>

Yuki2718:
I'll investigate, will add WIP back.

zloyden:
Just in case, this `/strp_v` script may not appear immediately, but after some page refresh.

Yuki2718:
Yes, it seems loaded quite conditionally via loadTool. I guess scriptlet can also be used, they may change path in future.

zloyden:
> Yes, it seems loaded quite conditionally via loadTool. I guess scriptlet can also be used, they may change path in future.

That's a good point. Please also add necessary domains to the our scriptlet rule.

Yuki2718:
Is `#%#//scriptlet("abort-on-property-read", "loadTool")` okay? I only found `#%#//scriptlet("abort-current-inline-script", "loadTool", "/'popping'/")` in common scriptlet filters but there's no popping in these cases. Or do you prefer "abort-current-inline-script" with a new third parameter?

Yuki2718:
Added a few domain. It's really not straightforward thx to their random redirect and its conditional nature, I'm almost giving up. Sorry but please add an according scriptlet on your side - all domains in /strp_v rule should be added. Also upon investigation I came across `https://www.novojoy.com/gorgeous-blonde-nicole-neal` and `https://www.novoporn.com/hannah` which require `#%#//scriptlet('abort-current-inline-script', 'document.querySelectorAll', 'popMagic')`.

zloyden:
> Is `#%#//scriptlet("abort-on-property-read", "loadTool")` okay? I only found `#%#//scriptlet("abort-current-inline-script", "loadTool", "/'popping'/")` in common scriptlet filters but there's no popping in these cases. Or do you prefer "abort-current-inline-script" with a new third parameter?

`#%#//scriptlet("abort-on-property-read", "loadTool")` is ok on my end or you can add another value to the old rule.

Yuki2718:
Where to add? It's the first time I add a new common scriptlet filter.

zloyden:
> Also upon investigation I came across `https://www.novojoy.com/gorgeous-blonde-nicole-neal` and `https://www.novoporn.com/hannah` which require `#%#//scriptlet('abort-current-inline-script', 'document.querySelectorAll', 'popMagic')`.

Added them, thanks.



> Where to add? It's the first time I add a new common scriptlet filter.

To the same place, I mean you can exchange this old rule with a new one (if it works for old domains) **or** just add a new scriptlet with domains that required.

zloyden:
I guess it's ok now, but I don't see these requests. Also you can make it via one rule using Regex.
```
/smallfr/*$subdocument,domain=babeimpact.com|decorativemodels.com|sexykittenporn.com
/smallfr2/*$subdocument,domain=babeimpact.com|decorativemodels.com
```

Yuki2718:
You can see here: `https://www.decorativemodels.com/galleries/hanna-looks-pretty-hazed-to-me/` but `babeimpact.com`  and `sexykittenporn.com` were only when navigated thorough redirect. Okay I'll make them a single rule - is regex preferred than `/smallfr*/*`?

zloyden:
> You can see here: `https://www.decorativemodels.com/galleries/hanna-looks-pretty-hazed-to-me/` but `babeimpact.com` and `sexykittenporn.com` were only when navigated thorough redirect. Okay I'll make them a single rule - is regex preferred than `/smallfr*/*$subdocument,domain=babeimpact.com|decorativemodels.com|sexykittenporn.com`?

I prefer Regex because it's more accurate than using such solutions, but in this case `/smallfr*/*$subdocument,domain=babeimpact.com|decorativemodels.com|sexykittenporn.com` is ok and one more point - it will be processed faster than regex. Could you also add `babeimpact.com,decorativemodels.com,sexykittenporn.com##iframe[src^="/smallfr"]`? it's is necessary for app.

Yuki2718:
Yeah, I thought so as in this case domain is limited - but for generic rule I'd prefer regex like you. Ok, a second.

zloyden (approved):

zloyden:
Thanks, nice work here:)

PR #70746, opened 2020-12-24
fix winsoft.ucoz.net #70707

gork7777:
#70707

### DIFF
--- a/RussianFilter/sections/specific.txt
+++ b/RussianFilter/sections/specific.txt
@@ -1552,6 +1552,7 @@ tlgg.ru##div[id^="ads"]
 serfery.ru,loxotrona.net##.b-adds
 ruspornovideo.su###secondary.widget-area
 ukr.net##div[data-side="left"][class*="fixed-block"] > div:nth-child(1) ~ div
+winsoft.ucoz.net##table:last-child > tbody > tr > td > table:nth-child(2) > tbody > tr > td > div.noforum > table > tbody > tr > td:nth-child(2) > table > tbody > tr:nth-child(2) > td:nth-child(2) > div:nth-child(8)
 ||zporno.porn/img/asd/
 rezka.ag###ownplayer-link-show
 golos.zp.ua##div[itemtype="https://schema.org/WPAdBlock"]


Yuki2718:
Maybe I could be more helpful. The reason the filter is problematic is, say, it doesn't match at all on my end. As it's too specific, any slight change on the site causes no match. It also consumes more memory. Make filters as simple as possible within a degree it won't cause false positive. Until you become intimate with filter writing, it'll be best to avoid nth-child or nth-of-type, as well as too long chain of >.

Here how to reach a right filter:

1. Press F12 and make sure Elements (Chrome) / Inspect (Firefox) is selected. If not, click ont the tab.

<details>

ucoz1

</details>

2. Click the upper-left icon and make sure it's color changed.

<details>

ucoz2

</details>

3. Hover over and click on the element you want to hide. Then DOM node of the element will be chosen on the tab. Alternatively, you can skip these steps and instead right-click on the element, then choose Inspect (Chrome) / Inspect Element (Firefox).

<details>
<summary>Screenshot 3</summary>

ucoz3

</details>

Now you see it's an a element whose href attribute points to `http://sub2.admitlead.ru/`, then it's just a matter of filter syntax. If you specify your filter well, it will catch both of fake buttons on the page and nothing else - there're some right filters and mine is just one option. In this case another candidate is a network filter `||ogosoft.ucoz.ru/4.png`. Often to use cosmetic filter or network filter is hard problem, but I'll choose cosmetic in this case as 4.png will likely to change soon.

gork7777:
Thank you very much for your nice help. So after finding this element, how do I know which filter to put it in?

Yuki2718:
@gork7777 That's a matter of experience. Filter authors try to not only block targets but also minimize likelihood of false positive and maximize efficiency/performance. Firstly read the document zloyden provided [here](https://github.com/AdguardTeam/AdguardFilters/pull/70866#issuecomment-751363172) and also [this](https://github.com/DandelionSprout/adfilt/blob/master/Wiki/SyntaxMeaningsThatAreActuallyHumanReadable.md) by @DandelionSprout. Next, you need to gain experience in the field. A reason we prefer `###div-Top-Leaderboard` to `##.col-md-12` in https://github.com/AdguardTeam/AdguardFilters/pull/70868 is that not only the former is descriptive  (what this element is is obvious by its name) but also we know col-md-12 is relatively common class name and thus false positive prone. There will be various ways to gain experience but in my case I tried to solve issues I encountered by myself. If I can't or am not confident, reported it so I can learn how experts solve it. I also used to comment possible fixes on open issues. Github notifies me once the issue was fixed and I can compare my fix with those of experts.

gork7777:
I'll try to learn with time. I'm trying to gain experience by looking at the problems solved. Thank you for your helps

gork7777:
One more thing. If there is a Phishing site, how do we solve it?

Yuki2718:
>  If there is a Phishing site, how do we solve it?

Just ping lolka4toli , he addresses all kinds of malicious sites with Browsing Security feature.

zloyden:
Guys, I really appreciate your help here, but one more thing - the website has low visitors. In this case, we don't add rules to our filters.

PR #70844, opened 2020-12-25
fix akdenizsonhavadis.com #70782

gork7777:
#70782


AdamWr:
I think that in this case it would be better to use something like:
`akdenizsonhavadis.com##.reklam_tikla`

Fixed in https://github.com/AdguardTeam/AdguardFilters/commit/c5d5d6f90b11b4a09a6f4994c0fa99acd56ab4e1

gork7777:
Yeah you're right, sorry for mistake

PR #70866, opened 2020-12-25
fix phoneoff.ru #70575

gork7777:
https://github.com/AdguardTeam/AdguardFilters/issues/70575

### DIFF
--- a/AnnoyancesFilter/sections/remains.txt
+++ b/AnnoyancesFilter/sections/remains.txt
@@ -75,7 +75,7 @@ link.tl##.fly_head_bottom
 extramovies.tv##.theiaStickySidebar > div[id^="ai_widget"]
 prizyv.ru###text-10.widget_text
 prizyv.ru###text-9.widget_text
-phoneoff.ru##aside#secondary.widget-area.is-active
+phoneoff.ru#?#.site-content > aside.widget-area:has(> div.widget_text > div.widget-header:contains(Реклама))
 xtreme.rip##.topBorder > table[border="0"][width="100%"][height="205"]+div.separator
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/remains.txt
```diff
@@ -75,6 +75,7 @@ link.tl##.fly_head_bottom
 extramovies.tv##.theiaStickySidebar > div[id^="ai_widget"]
 prizyv.ru###text-10.widget_text
 prizyv.ru###text-9.widget_text
+phoneoff.ru##aside#secondary.widget-area.is-active
```
zloyden, line 78: Use ext-css here. `phoneoff.ru#?#.site-content > aside.widget-area:has(> div.widget_text > div.widget-header:contains(Реклама))`

zloyden (changes requested):

zloyden (approved):

zloyden:
Ok, good work here:)

gork7777:
Can you help me little for how to do css for filters?

zloyden:
Of course, you can read about creating rules using a pseudo-class [here](https://kb.adguard.com/en/general/how-to-create-your-own-ad-filters#extended-css-has). There is a lot of useful information.
It is also useful to look at the current filter database in the section `css_extended.txt` - there you can see what basic principles we apply when creating such rules.

The main purpose of using such rules is when basic rules are not enough. In your case, your rule is almost fine, but it's pretty tricky. To avoid false positives and make the rule more accurate, it is safer to use ext-css.

gork7777:
Oh... it'll take me time to learn, it's too complicated

PR #70868, opened 2020-12-25
fix tribunjualbeli.com #70799

gork7777:


Yuki2718:
I'd prefer `tribunjualbeli.com###div-Top-Leaderboard` in this case. At least you have to narrow it down if you use such a common class name, like `tribunjualbeli.com##.col-md-12[style="margin-bottom: 20px;"][align="center"]`. You can drop the element name div though.

AdamWr:
>I'd prefer `tribunjualbeli.com###div-Top-Leaderboard` in this case.

Me too.
Added in - https://github.com/AdguardTeam/AdguardFilters/commit/49028ca6ce14f5eb1808739a37018886b8a92dcf

PR #71060, opened 2020-12-28
fix #70536 inc42.com

### LINKED ISSUE #70536: inc42.com
 [https://inc42.com/buzz/facebooks-double-standards-questioned-as-kisan-ekta-morcha-page-taken-down-temporarily/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Finc42.com%2Fbuzz%2Ffacebooks-double-standards-questioned-as-kisan-ekta-morcha-page-taken-down-temporarily%2F)

### Screenshots
<details>

  Screenshot 1
</details>
<details>

  Screenshot 2
</details>

### System configuration

Information | value
--- | ---
AdGuard product: | AdGuard Browser extension v3.5.25
Browser: | Firefox
Stealth mode options: | Hide your search queries, <br/>Send Do-Not-Track header, <br/>Strip URLs from tracking parameters, <br/>Self-destructing third-party cookies (2880), <br/>Block WebRTC, <br/>Hide your Referrer from third-parties
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base, <br/>Peter Lowe's Blocklist<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection, <br/>EasyPrivacy, <br/>Fanboy's Enhanced Tracking List, <br/>Fanboy Anti-Facebook List<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> AdGuard Annoyances, <br/>Fanboy's Annoyances, <br/>Web Annoyances Ultralist, <br/>Adblock Warning Removal List, <br/>I don't care about cookies<br/><br/><b>Security:</b><br/> Online Malicious URL Blocklist

stanislav-atr:
#70536

### DIFF
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -116,6 +116,7 @@ inc42.com#$#iframe[id^="offer-"] { display: none !important; }
 inc42.com#$#.tp-backdrop { display: none !important; }
 inc42.com#$#.tp-modal { pointer-events: none !important; overflow: hidden !important; }
 inc42.com#$#iframe[id^="piano-id-"] { pointer-events: all !important; }
+inc42.com#$#html > body > div.plus-bottom-bar[style] { display: none !important; }
 inc42.com#$#button.tp-close { pointer-events: all !important; }
 inc42.com#$?#body.tp-modal-open { overflow: visible !important; }
 inc42.com#$#body { overflow: visible !important; }
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -112,7 +112,7 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
 bankbazaar.com#$#.modal-backdrop { display: none !important; }
 bankbazaar.com#$##js-sign-in-box[style="display: block;"] ~ .modal-backdrop { display: block !important; }
 magadanmedia.ru###social_modalblock
-inc42.com#$#.tp-modal { display: none !important; }
+inc42.com#$#iframe[id^="offer-"] { display: none !important }
 inc42.com#$#.tp-backdrop { display: none !important; }
 inc42.com#$#body { overflow: visible !important; }
 rockstargames.su##.webnavoz_notificationbox
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -112,8 +112,12 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
 bankbazaar.com#$#.modal-backdrop { display: none !important; }
 bankbazaar.com#$##js-sign-in-box[style="display: block;"] ~ .modal-backdrop { display: block !important; }
 magadanmedia.ru###social_modalblock
-inc42.com#$#iframe[id^="offer-"] { display: none !important }
+inc42.com#$#iframe[id^="offer-"] { display: none !important; }
 inc42.com#$#.tp-backdrop { display: none !important; }
+inc42.com#$#.tp-modal { pointer-events: none !important; overflow: hidden !important; }
+inc42.com#$#iframe[id^="piano-id-"] { pointer-events: all !important; }
+inc42.com#$#button.tp-close { pointer-events: all !important; }
+inc42.com#$?#body.tp-modal-open { overflow: visible !important; }
 inc42.com#$#body { overflow: visible !important; }
 rockstargames.su##.webnavoz_notificationbox
 zerohedge.com###exitModalOverlay
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -112,6 +112,9 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
 bankbazaar.com#$#.modal-backdrop { display: none !important; }
 bankbazaar.com#$##js-sign-in-box[style="display: block;"] ~ .modal-backdrop { display: block !important; }
 magadanmedia.ru###social_modalblock
+inc42.com#$#.tp-modal { display: none !important; }
+inc42.com#$#.tp-backdrop { display: none !important; }
+inc42.com#$#body { overflow: visible !important; }
 rockstargames.su##.webnavoz_notificationbox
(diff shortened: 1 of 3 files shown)


AnnoyancesFilter/sections/popups.txt
```diff
@@ -112,6 +112,9 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
 bankbazaar.com#$#.modal-backdrop { display: none !important; }
 bankbazaar.com#$##js-sign-in-box[style="display: block;"] ~ .modal-backdrop { display: block !important; }
 magadanmedia.ru###social_modalblock
+inc42.com#$#.tp-modal { display: none !important; }
```
AdamWr, line 115: It seems that it blocks a login popup.

image

</details>

EnglishFilter/sections/whitelist.txt
```diff
@@ -42,6 +42,10 @@ youtube.com#@##header.ytd-c4-tabbed-header-renderer:style(position: absolute !im
 ! https://github.com/AdguardTeam/PopupBlocker/issues/215
 @@||senpa.io^$extension
 @@||vanis.io^$extension
+! https://github.com/AdguardTeam/AdguardFilters/issues/70536
+inc42.com#@#.tag-social
+@@||piano.io^$domain=inc42.com
+@@||tinypass.com^$domain=inc42.com
```
AdamWr, line 48: As far as I understand `inc42.com#@#.tag-social` is needed, because it hides content and it's from Fanboy's Social Blocking List/Fanboy's Annoyances.
What about these two:
```
@@||piano.io^$domain=inc42.com
@@||tinypass.com^$domain=inc42.com
```
As far as I see `tinypass.com` is blocked by `Fanboy's Enhanced Tracking List`, and it looks like that it breaks a website, so fine. But could you please tell me why `@@||piano.io^$domain=inc42.com` is needed?

AnnoyancesFilter/sections/popups.txt
```diff
@@ -112,6 +112,9 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
 bankbazaar.com#$#.modal-backdrop { display: none !important; }
 bankbazaar.com#$##js-sign-in-box[style="display: block;"] ~ .modal-backdrop { display: block !important; }
 magadanmedia.ru###social_modalblock
+inc42.com#$#iframe[id^="offer-"] { display: none !important }
```
AdamWr, line 115: Please re-add `;` after `!important`.
Other problem is that, when popup is displayed, content is not clickable due to `tp-modal` element.

AnnoyancesFilter/sections/popups.txt
```diff
@@ -112,6 +112,9 @@ bankbazaar.com#$#.oneTapLoginModal[style] { display: none !important; }
...
+inc42.com#$#.tp-backdrop { display: none !important; }
+inc42.com#$#body { overflow: visible !important; }
```
AdamWr, line 117: It seems that it doesn't work with apps, because there is `!important` already and it has higher priority.
Please change it to something like:
`inc42.com#$#body.tp-modal-open { overflow: visible !important; }`

EnglishFilter/sections/whitelist.txt
stanislav-atr: With piano.com^ blocked:

image
</details><br/>
And login button is disabled.

stanislav-atr:
The problem is that the same script from `piano.com^` does login screen and popup, wich both use the same modal element.

EnglishFilter/sections/whitelist.txt
AdamWr: Which filter list block `piano.com`?

AnnoyancesFilter/sections/popups.txt
stanislav-atr: Could you please try 
`inc42.com#$#.tp-modal { pointer-events: none !important; }
`
on your side?

EnglishFilter/sections/whitelist.txt
stanislav-atr: Peter Lowe's Blocklist

AdamWr: Ok then, but I guess that we should also add to `Tracking Protection filter`:
`||piano.io/tracker/$domain=inc42.com,important`

stanislav-atr:
Right now i have everything working with these:
```
inc42.com#@#.tag-social
@@||piano.io^$domain=inc42.com
@@||tinypass.com^$domain=inc42.com
!
inc42.com#$#iframe[id^="offer-"] { display: none !important }
inc42.com#$#.tp-backdrop { display: none !important; }
inc42.com#$#.tp-modal { pointer-events: none !important; }
inc42.com#$#iframe[id^="piano-id-"] { pointer-events: all !important; }
inc42.com#$#button.tp-close { pointer-events: all !important; }
inc42.com#$?#body.tp-modal-open { overflow: visible !important; }
inc42.com#$#body { overflow: visible !important; }
```

AnnoyancesFilter/sections/popups.txt
AdamWr: It works, but if I'm not wrong, it causes that login popup is not clickable, I mean that for example you can't close it.

stanislav-atr: yeah, just fixed it a minute ago with `pointer-events: all` where needed

AdamWr:
>i have everything working with these right now:

It looks fine now, but one more issue is that I have two scrollbars and there is one annoying element at the bottom.

image

</details>

Scrollbar can be fixed by for example changing `inc42.com#$#.tp-modal { pointer-events: none !important; }` to:
`inc42.com#$#.tp-modal { pointer-events: none !important; overflow: hidden !important; }`

For this element at the bottom:
`inc42.com##html > body > div.plus-bottom-bar[style]`

stanislav-atr:
> For this element at the bottom:
> `inc42.com##html > body > div.plus-bottom-bar[style]`

Could not reproduce this one on my devices.

AdamWr:
>Could not reproduce this one on my devices, is that ad leftover?

I see it only on the first visit and it's leftover after blocking `iframe[id^="offer-"]`, so I think it should be added to Annoyances filter.

AdamWr (approved):

PR #71128, opened 2020-12-28
tribundergi.com

gork7777:


TurkishFilter/sections/whitelist.txt
```diff
@@ -1,4 +1,6 @@
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/71125
+@@||platform.twitter.com^$domain=tribundergi.com
```
zloyden, line 3: Please remove domain from the common rule which starts with `||platform.twitter.com/widgets.js$domain=` instead of the exception.

zloyden (changes requested):

zloyden:
Fixed it directly.

PR #71173, opened 2020-12-29
Trackers: tapas.io

ghost:
### URL: `https://tapas.io/episode/1051887`
Screenshot:
Screenshot

### Network Logs
network1
network2


SpywareFilter/sections/specific.txt
```diff
@@ -22,6 +22,7 @@
 ||nd.demdex.net/event?d_mid=$redirect=nooptext,important,domain=9now.com.au
 ||adc.nine.com.au/?token=$redirect=nooptext,important,domain=9now.com.au
 !
+||paypal.com/xoplatform/logger/
```
Alex-302, line 25: This rule is dangerous.

Alex-302: Added as domain specific.

piquark6046: > 
> 
> This rule is dangerous.

Did a rule like that make a problem?

PR #71315, opened 2020-12-31
Update #71217 (remove old domain)

Yuki2718:
As in the issue comment, the old domain redirects to the new one so probably can be removed.


AdamWr (approved):

AdamWr:
Thank you.

PR #71710, opened 2021-01-05
kointimes.net

gork7777:


AdamWr:
It seems that it has been fixed in this commit -https://github.com/AdguardTeam/AdguardFilters/commit/59b8b15eb15a27b18a1ab28ce559b1270f4f370a

PR #71983, opened 2021-01-10
#71968 withukor.com

wout-codes:
#71968


Alex-302 (approved):

Alex-302:
Also `generichide` exclusion is required for platforms without JS rules support: https://github.com/AdguardTeam/AdguardFilters/commit/f8531da5df63bd96895144b742f64ee04282982b

PR #72039, opened 2021-01-11
Fix/#71696 bitfly.io

stanislav-atr:
#71696


zloyden:
https://github.com/AdguardTeam/AdguardFilters/pull/72099

PR #72351, opened 2021-01-16
fix haberturk.com ads head

muhasturk:


Sergey-Lyapin:
I see that all `.ads-zone` elements are empty, please specify an url where this element located.

gork7777:
Specified site: www.haberturk.com
`##.ads-zone` is already blocked in the filtering log.

PR #72586, opened 2021-01-18
fix #72583 mega.nz

### LINKED ISSUE #72583: mega.nz
### Issue URL (Annoyance)
 [https://mega.nz/file/lsQywayI#2H8X3uUhiI6qPStE93xeHGIYAP5rVycXjFVR3G3FeR4](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fmega.nz%2Ffile%2FlsQywayI%232H8X3uUhiI6qPStE93xeHGIYAP5rVycXjFVR3G3FeR4)
### Comment
Username: @Rakleed
> Cookie
> --
> Информация ниже сгенерирована автоматически. Не удаляйте ее.
> 
> Referrer:
> https://www.car-auto-repair.com/
> 
> Custom userscripts:
> 

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
Browser: | Chrome
AdGuard driver: | WFP
Stealth mode options: | Hide your search queries, <br/>Remove X-Client-Data header from HTTP requests, <br/>Strip URLs from tracking parameters, <br/>Self-destructing third-party cookies (180), <br/>Hide your Referrer from third-parties (mega.nz), <br/>Hide your User-Agent ()
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> AdGuard Annoyances<br/><br/><b>Language-specific:</b><br/> AdGuard Russian

Yuki2718:
https://github.com/AdguardTeam/AdguardFilters/issues/72583
CSS filter is optional, even without it the site displays normally after some time so I didn't made the rest of rules to `#$#`.


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -87,6 +87,8 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 ||cdn.privacy-mgmt.com^$important,domain=promipool.de|android-hilfe.de
 !
 !
+mega.nz##.fm-dialog-overlay
+mega.nz#$#.overlayed .scroll-block { filter: none!important;-webkit-filter: none!important; }
```
slavaleleka, line 91: for such cases rules should be type alike, so both should be `#$#`

also space is missed in second rule before `-webkit...`

slavaleleka (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
Yuki2718: > for such cases rules should be type alike, so both should be `#$#`

I understand and thus commented, you mean still I should? Then Content blocker etc. will see the cookie consent.

> also space is missed in second rule before `-webkit...`

okay, will change.

Yuki2718: I mean, we can even remove css filer if you like.

ghost (approved):

AnnoyancesFilter/sections/cookies_specific.txt
zloyden: Hm, mobile view is ok with those rules. If we use `#$#` for `.fm-dialog-overlay`, it won't work for content blockers.

zloyden (approved):

PR #72839, opened 2021-01-21
Issue labeler

slavaleleka:
there is an issue with `A: Resolved` label which will be removed if filters issue was closed and now is opening (usually it happens after newer cases reported) https://github.com/AdguardTeam/ReportsWebApp/issues/112
the same will happen with `A: Cannot reproduce`

~~also I noticed that a lot of closed issues are marked by `A: In progress`. I assume the label is just forgotten and it is not left on purpose so I've made it's auto-removing on issue close~~

~~I suggest to mark issue by `A: In progress` label when it got assigned~~


Alex-302:
2 - I use it for checking user's feedbacks) And remove this label regularly.

ameshkov (approved):

Alex-302 (changes requested):
https://github.com/AdguardTeam/AdguardFilters/pull/72839#issuecomment-764799527

Sergey-Lyapin:
3. I think it is better to add the label manually to know when the de facto work started.

Alex-302 (approved):

PR #73203, opened 2021-01-25
Kabum.com.br and others

fabianski7:

image
</details>

This is a recent tracker and is growing in Brazil since the approval of LGPD (Brazilian version of GDPR)

You can find it on several local sites. Some examples in addition to the already mentioned:

<details><summary>meioemensagem.com.br</summary>

image
</details>

<details><summary>piracanjuba.com.br</summary>

image
</details>


fabianski7:
this rule will also block the cookie popup that appears on image 2

Alex-302:
In this case goadopt.io should be blocked in Tracking Protection and Annoyances filter.
Also blocked `tag.hariken.co`

PR #73356, opened 2021-01-27
walmart.ca

gork7777:


zloyden (approved):

zloyden:
Ok, I guess it's stable for now, but I will add another rule.

PR #73389, opened 2021-01-27
zmega.com

gork7777:


zloyden (approved):

zloyden:
@gork7777 I've changed the rule a bit, because if we use a cosmetic one, you'll see a leftover. Please notice a difference and thanks for participation:)

PR #73391, opened 2021-01-27
plus.diolinux.com.br

fabianski7:


Alex-302:
Not looks as analytics
image

fabianski7:
See how much POST the script performed during the time it was open. I followed them and it seems to me to be a tracker that informs on which page there are active users and their activity time.

fabianski7:
sometimes it does not send anything, so the difference in size between the packets

PR #73460, opened 2021-01-28
elu24.ee

gork7777:


zloyden (approved):

zloyden:
I removed one rule, because it's already on the Estonian List.

PR #73469, opened 2021-01-29
more kicka** domains

wout-codes:


zloyden (changes requested):
`katfreak.*,kickasshydra.*,katbay.*` please also add specific domains without `*` to the end of each list. It's useful for Safari.

wout-codes:
Which line?

wout-codes:
By the way, there are ads in the Kicka**. Pop-ups are blocked.

zloyden (approved):

zloyden:
> I can't add all the domains. There are like over 50 domains. I only know one.

Ok, approved.


> there are ads in the Kicka**

I'll change it manually, but we really should wait for a while.

wout-codes:
Okay. This website has pop-ups with AdGuard for Windows `https://shortlinkpay.com/276a8FW`. With the AdGuard Browser Extension it's blocked. Can you figure the problem? Is it a bug in AdGuard Windows?

zloyden:
Fixed, it should be fine for the app. But after fixing it I don't have access to this site)

wout-codes:
> Fixed, it should be fine for the app. But after fixing it I don't have access to this site)

What's the problem with the app? Doesn't it support any syntax?

zloyden:
Tbh, I don't know. The website is blocked on my end. I think an additional rule was needed because of a different approach to filtering the page.

PR #73537, opened 2021-01-29
Fix/#72436 rtl.de

stanislav-atr:
#72436


AdamWr:
Extensions for Firefox and Opera can't update/download new JavaScript rules, so please remove `rtl.de` from these rules:
https://github.com/AdguardTeam/AdguardFilters/blob/8fae8684d84b998e3673b0ba42115c857a10db14/AnnoyancesFilter/sections/cookies_specific.txt#L80-L81
and add them with `!#if` directive, for example:
```
!#if (adguard_app_windows || adguard_app_mac || adguard_app_android || adguard_ext_chromium)
rtl.de#$#div[id^="sp_message_"] { display: none !important; }
rtl.de#$#html.sp-message-open { overflow: visible !important; position: static !important; margin-top: 0!important; }
!#endif
```

AdamWr (approved):

PR #73760, opened 2021-02-01
udc.yahoo.com

fabianski7:


ghost (approved):

Alex-302:
Not sure about it. It seems it used for performance check.

fabianski7:
> Not sure about it. It seems it used for performance check.

So what? That increases entropy.

nvm. fuck

Alex-302:
They can recognize and track you even without this check.

PR #73840, opened 2021-02-02
Update general_extensions.txt

wout-codes:
More issues.


wout-codes:
@zloyden More proofs found for AdGuard App, which can be fixed by this rule `$$script[tag-content="zoneSett"]`

Alex-302:
no need to add dozens of comments. Better edit existing comment on github.

zloyden:
I agree. 3 comments is enough (in this case, this is rather an exception to the rule). Generally, please do as @Alex-302 said `Better edit existing comment on github.`

PR #74096, opened 2021-02-05
Fix/#72558 mojanorwegia.pl

stanislav-atr:
#72558


AdamWr:
Could you please also block this ad leftover:

image

</details>

I think that something like this should be fine:
`mojanorwegia.pl##.newArticleA`

Please also add rules to hide share buttons:
```
mojanorwegia.pl##.right-col > .fb
mojanorwegia.pl##.newshare-wrapper
```

zloyden (approved):

PR #74202, opened 2021-02-06
dunya.com

gork7777:


Yuki2718:
@gork7777 Off-topic but are you aware of or told about AdGuard tab on devtools? If not, it can potentially save your time to create filters.

gork7777:
Yeah I knew this AdGuard tab on devtools

Issue #74247, opened 2021-02-07
diakov.net


ObsessedHacker:
[//]: # (***You can delete or ignore strings starting with "[//]:" They will not be visible either way.)

***Description***: diakov.net - Incorrect blocking
* **Current behaviour**: Blocking "Subscribe" button

[//]: # (Substitute this line with a description of the problem)

[//]: # (Replace %screenshot_url% below with a link to the screenshot of the problem. Also, you can paste image from clipboard instead. It will be automatically loaded.)
<details><summary>Screenshot:</summary>

![image](https://user-images.githubusercontent.com/13978119/107133175-17754a80-68ee-11eb-9fdf-0d8d8a976e35.png)

</details><br/>

* **Expected behaviour**: Showing the "Subscribe" button

[//]: # (Substitute this line with a description of what should happen normally. If needed, provide a screenshot below, same as above)

<details><summary>Screenshot:</summary>

![image](https://user-images.githubusercontent.com/13978119/107133187-3f64ae00-68ee-11eb-8215-a832f92921f3.png)
</details><br/>

***Steps to reproduce the problem***: https://diakov.net/12031-wondershare-filmora-x-1012016.html

[//]: # (Substitute this line with a step-by-step instruction how to reproduce the issue)

***System configuration***

**Filters:**

[//]: # (Substitute this line with the list of your active filters, separated by commas)

[//]: # (Please enter the correct values for your case to the table below)

Information                            | Value
---                                    | ---
Operating system:                      | Windows 10 Pro 64bit 20H2 19042.746
Browser:                               | Google Chrome Version 88.0.4324.150 (Official Build) (64-bit)
AdGuard version:                       | uBlock Origin v1.32.4
Filters enabled:                       | AdGuard Annoyances

[//]: # (This template is meant for missed ad/false positive reports, for other type of reports edit it accordingly)
[//]: # (If this is a crash report, include the crashlog with https://gist.github.com/)

PR #74248, opened 2021-02-07
Show "Subscribe" button on diakov.net

Obegg:
Issue reference: diakov.net #74247


Alex-302:
This is expected behavior, if AdGuard Annoyances filter is enabled.

ObsessedHacker:
> This is expected behavior, if AdGuard Annoyances filter is enabled.

It's incorrect blocking, the subscribe button should be shown. It's not an annoyance.
Once you subscribe you receive email notifications for updates.

Alex-302:
In AdGuard Annoyances filter we may block everything. It is not enabled by default.

Alex-302:
You can add it by parts:
https://github.com/AdguardTeam/AdguardFilters/tree/master/AnnoyancesFilter/sections
Without `annoyances.txt` in your case.

PR #74563, opened 2021-02-11
missed ad server

wout-codes:
Serves fake/scam notifications in top right corner.

### DIFF
--- a/EnglishFilter/sections/adservers.txt
+++ b/EnglishFilter/sections/adservers.txt
@@ -5,6 +5,7 @@
 !
 !! https://a.bestcontentitem.top/warp/*?r=
 !
+||zopim.com^$third-party
 ||loopoohy.com^
 ||continuedhostilityequipped.com^$document,popup
 ||jcqwzssqks.com^


EnglishFilter/sections/adservers.txt
```diff
@@ -5,6 +5,7 @@
 !
 !! https://a.bestcontentitem.top/warp/*?r=
 !
+||zopim.com^$third-party
```
zloyden, line 8: We have already it in the Annoyances filter.

wout-codes: @zloyden Don't make it third party.

zloyden: What do you mean? As I said earlier, it's in the Annoyances filter https://github.com/AdguardTeam/AdguardFilters/blob/master/AnnoyancesFilter/sections/widgets.txt#L176

wout-codes: @zloyden Like this: `||zopim.com^$third-party` to this `||zopim.com^` without any domains.

wout-codes: Ok wait nevermind, it may cause false positives

zloyden: Yep, that's why we added the domains with `~`. In other words, these are exceptions.

wout-codes: Yes.

PR #74598, opened 2021-02-11
help.keenetic.com

gork7777:


AnnoyancesFilter/sections/annoyances.txt
```diff
@@ -2403,6 +2403,7 @@ b-g.by##.ban-bottom-post
 pronetblog.by#?#aside > section.widget-after-post-content:has(> div.textwidget > div[id] > div.rss2)
 stroi.mos.ru###subscribe-title
 osp.ru#?#.right-column > div.panel:has(> div.panel-body > script[src^="https://email.osp.ru/"])
+help.keenetic.com###formsubscribe
```
zloyden, line 2406: I don't think that we should block it in the footer.

PR #74673, opened 2021-02-12
pop-ups at 9anime.* sites

wout-codes:
Happens with only AdGuard App.


AdamWr:
I can't reproduce popups on `9anime.to`, but I guess that may happens on `9anime.at`.
`window.open` can be used not only for ads/popups, so `prevent-window-open` may breaks some website functionality.
In case of this website I prefer more specific rule (unless they will start changing their popups script very often), so could you please check if these rules work or `9anime.at` (these rules work fine on my end):
https://github.com/AdguardTeam/AdguardFilters/blob/dcaa6b98d68325a71ddacdba86d4a8d71203d681/EnglishFilter/sections/general_extensions.txt#L429-L430
and if it will works then change `9anime.ru,9anime.to` to `9anime.*`?

By the way, if you are creating pull request not related to opened issue then please provide example link.

wout-codes:
> I can't reproduce popups on `9anime.to`, but I guess that may happens on `9anime.at`.
> `window.open` can be used not only for ads/popups, so `prevent-window-open` may breaks some website functionality.
> In case of this website I prefer more specific rule (unless they will start changing their popups script very often), so could you please check if these rules work or `9anime.at` (these rules work fine on my end):
> https://github.com/AdguardTeam/AdguardFilters/blob/dcaa6b98d68325a71ddacdba86d4a8d71203d681/EnglishFilter/sections/general_extensions.txt#L429-L430
> 
> 
> and if it will works then change `9anime.ru,9anime.to` to `9anime.*`?
> By the way, if you are creating pull request not related to opened issue then please provide example link.

Works! Let me change

AdamWr:
Thank you.

PR #74700, opened 2021-02-12
placeholder/leftover

wout-codes:
Placeholder/leftover at `yoshare.net`.


AdamWr:
It seems that there is another ad leftover - `https://yoshare.net/garena-free-fire-mod-apk/`

image

</details>

What about adding something like that:
`yoshare.net##iframe[id^="yoshare.net_"][id$="_DFP"]`

wout-codes:
Ok

AdamWr (approved):

AdamWr:
Thank you.

PR #75046, opened 2021-02-16
grouped and added a domain

wout-codes:
Added a domain.


Sergey-Lyapin:
I did it earlier https://github.com/AdguardTeam/AdguardFilters/commit/dd9918f54d56c8f753f4a091afbe508002318d48
Please don't forget to add a link to the issue in the pull request description.

wout-codes:
Oh so there was an issue? I didn't see it. By the way, I visit that site myself too that's how I found it out.

PR #75256, opened 2021-02-19
free.drweb.ru

gork7777:


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -294,7 +294,7 @@ infranken.de#$#div[id^="cmpbox"] { display: none !important; }
 skillbox.ru##app-cookies
 ||bratpfannentest.de/wp-content/plugins/borlabs-cookie^
 nordbayern.de##div[id^="cmpbox"]
-drweb.com###cookie_accept_msg_block
+free.drweb.ru,drweb.com###cookie_accept_msg_block
```
zloyden, line 297: This is necessary for `drweb.ru`.

zloyden (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
Yuki2718: @zloyden You mean false positive? BTW the same cookie consent is present on all locales e.g. `free.drweb.co.jp`.

Yuki2718: And even wildcard is not enough, German version is `drweb-av.de`

zloyden: I meant that we don't need to use subdomain `free.` because it's needed for `drweb.ru`.
There is no need for wildcard, just use their domains.

zloyden: Changed the rule here.

zloyden (approved):

PR #75411, opened 2021-02-20
Fix naver.com

ghost:
#75386
#75385


AdamWr:
@piquark6046 Could you please move this rules to `foreign.txt` and add themo above this rule:
https://github.com/AdguardTeam/AdguardFilters/blob/ba99129253733ed83464dbc4696f290d0831d89e/EnglishFilter/sections/foreign.txt#L5533-L5534

piquark6046:
@AdamWr Is it okay?

AdamWr (approved):

Yuki2718:
I'll open PR in EL so remove once fixed in upstream.

PR #75420, opened 2021-02-20
Fixes some Anti-AdBlocks and popups

wout-codes:
Fixes some Anti-AdBlocks and pop-ups from several sites.
Use AdGuard for Windows, Mac or Android to reproduce popups and Anti-AdBlock

Links:
1. `https://coindice.win/`                                    -         Anti-AdBlock
2. `https://freebitcoin.win/home`                       -         shortly after you will be redirected to disable AdGuard (requires an account).
3. `https://cut-fly.com/OHjzi` - Popups and Anti-AdBlock
4. `https://www.gioialive.it/`   - Anti-AdBlock
5. `https://hanimesubth.com/` - Popups
6. `https://mega4up.com/lr0mwpg7q0va`        -          pop-ups at the last page after Captcha.
7. `https://diariodegoias.com.br/goiania-tem-mais-um-caso-de-falha-na-aplicacao-da-vacina-veja/` - Anti-AdBlock
8. `guideautoweb.com` - Anti-AdBlock
9. `burningseries.co` - Popups
10. `bs.to` - Popups
11. `https://searchenginereports.net/`   -  Placeholders/leftovers
12. `dailyvideoreports.net` - Anti-AdBlock
13. `https://eroticmv.com/young-nurses-in-love-1984/?asgtbndr=1` - Pop-ups
14. `https://readmanga.se/manga/the-beginning-after-the-end/` - Useless Ad-Server and Anti-AdBlock
15. `https://courses4u.info/` - Anti-AdBlock
16. `https://www.friendproject.net/home.php` - Anti-AdBlock (requires an account)

PR #75470, opened 2021-02-21
fix #75372 lacalleochotv.org

### LINKED ISSUE #75372: lacalleochotv.org
### Issue URL (Anti Adblock Script)
 [http://lacalleochotv.org/canal-4-online-en-directo-gratis/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=http%3A%2F%2Flacalleochotv.org%2Fcanal-4-online-en-directo-gratis%2F)

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
AdGuard product: | AdGuard for Windows v7.6 nightly 18
Browser: | Opera
AdGuard driver: | WFP
Stealth mode options: | Hide your search queries, <br>Send Do-Not-Track header, <br>Remove X-Client-Data header from HTTP requests, <br>Strip URLs from tracking parameters, <br>Self-destructing third-party cookies (180), <br>Block Push API, <br>Block Location API
DNS filtering: | server: `https://dns-unfiltered.adguard.com/dns-query`<br>filters: `User rules`
Filters: | <b>Ad Blocking:</b><br/> EasyList, <br/>Peter Lowe's Blocklist, <br/>AdGuard Mobile Ads, <br/>AdGuard Base<br/><br/><b>Privacy:</b><br/> EasyPrivacy, <br/>AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/> AdGuard Social Media<br/><br/><b>Annoyances:</b><br/> Adblock Warning Removal List, <br/>I don't care about cookies, <br/>EasyList Cookie List, <br/>AdGuard Annoyances<br/><br/><b>Security:</b><br/> Online Malicious URL Blocklist, <br/>Spam404, <br/>NoCoin Filter List<br/><br/><b>Language-specific:</b><br/> AdGuard Spanish/Portuguese
Custom filter subscriptions: | https://adguardteam.github.io/AdGuardSDNSFilter/Filters/filter.txt

wout-codes:
#75372

### DIFF
--- a/SpanishFilter/sections/antiadblock.txt
+++ b/SpanishFilter/sections/antiadblock.txt
@@ -18,8 +18,8 @@ forum.lfspro.net#@#.adsbygoogle
 @@||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$domain=forum.lfspro.net
 !
 ! https://github.com/AdguardTeam/AdguardFilters/issues/75372
+! https://github.com/AdguardTeam/AdguardFilters/issues/72368
 lacalleochotv.org#%#//scriptlet("abort-current-inline-script", "document.getElementById", "block")
-!+ PLATFORM(ios, ext_android_cb, ext_safari)
 @@||vergranhermano.com/player/ads.js
 ! https://github.com/AdguardTeam/AdguardFilters/issues/75110
 todopolicia.com#%#//scriptlet("abort-current-inline-script", "$", "initDetection")
@@ -31,8 +31,6 @@ deportealdia.live#%#//scriptlet("set-constant", "adblockDetector.init", "noopFun
 @@||mahimeta.com/networks/ad_code.php$domain=1f1.in|1i1.in
 ||mahimeta.com/networks/ad_code.php$xmlhttprequest,redirect=nooptext
 1f1.in,1i1.in#%#//scriptlet("abort-on-property-read", "mMCheckAgainBlock")
-! https://github.com/AdguardTeam/AdguardFilters/issues/72368
-@@||vergranhermano.com/player/ads.js$domain=adictosalatele.com
 ! https://github.com/AdguardTeam/AdguardFilters/issues/71820
 @@||forum.outerspace.com.br/js/siropu/am/ads.min.js
 ! https://github.com/AdguardTeam/AdguardFilters/issues/71853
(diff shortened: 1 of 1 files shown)


AdamWr (approved):

AdamWr:
If blocked request is just a bait and it's used on other websites as well, then I think it can be added for all platforms.

wout-codes:
@AdamWr Can you tell me how to use this `!+ PLATFORM(ios, ext_android_cb, ext_safari)`? and when can I use. If I don't use it what should the order be? Scriptlet first or whitelisting files to kill anti-adblock?

AdamWr:
This hint "`!+ PLATFORM(ios, ext_android_cb, ext_safari)`" means that rule will be only available in filter list for AdGuard for iOS, AdGuard Content Blocker for Android and AdGuard Extensions for Safari.

>when can I use it?

For example, in case if you need to unblock ad request to fix anti-adblock script, for platforms which do not support scriptlets and redirect rules.
If you unblock only a bait then in my opinion it can be added for all platforms.

>If I don't use it what should the order be? Scriptlet first or whitelisting files to kill anti-adblock?

It depends on the case, on `lacalleochotv.org` both rules will be applied.

wout-codes:
> It depends on the case, on lacalleochotv.org both rules will be applied.
@AdamWr Will the whitelisted rule get injected first in platforms with Scriptlet support or the scriptlet will get injected first in platforms with Scriptlet support?

AdamWr:
In this case anti-adblock will be fixed by the scriptlet, but the request will be unblocked by the whitelisted rule anyway.

wout-codes:
@AdamWr 
Then how about for this?
`example.com#@#.ad-top`
`example.com#%#//scriptlet("set-constant", "adblock", "false")`
Same case?

AdamWr:
`example.com#@#.ad-top` doesn't unblock any request, it unhides hidden element.
Better use a real case, because it may depends on how anti-adblock script looks like.

wout-codes:
> `example.com#@#.ad-top` doesn't unblock any request, it unhides hidden element.
> Better use a real case, because it may depends on how anti-adblock script looks like.

@AdamWr 
It's from my PR (https://github.com/AdguardTeam/AdguardFilters/pull/75420), also, to reproduce it requires an account. 
Here is the fix I could write: 
```
friendproject.net#@#.ad-top
friendproject.net#%#//scriptlet("set-constant", "adblock", "false")
```

AdamWr:
So I would have to check it to answer, but if it doesn't unblock ads, then I think it's fine to add it for all platforms without hints.

PR #75889, opened 2021-02-26
fix #75857 lookmovie.io

### LINKED ISSUE #75857: lookmovie.io [Pop-Up]
https://lookmovie.io/

Popups seen on the page when clicking on the search icon. Checked with Chrome browser on Android. Please fix

wout-codes:
#75857
They changed their code a little which is causing pop-ups.

### DIFF
--- a/EnglishFilter/sections/general_extensions.txt
+++ b/EnglishFilter/sections/general_extensions.txt
@@ -232,6 +232,8 @@ kat.*,kickass.*,kickass2.*,kickasstorrents.*,kat2.*,kattracker.*,thekat.*,thekic
 !
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/75857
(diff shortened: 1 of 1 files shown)


wout-codes:
@zloyden Can I craft rules like this, or is there a common rule? If there is I couldn't find.

zloyden:
We have common rules for this popups, but I'll change it directly.

wout-codes:
None of them works I think. The current common rule doesn't work. They changed their code.

wout-codes:
Try it yourself with AdGuard for Windows.

zloyden:
I already checked it with our products.


> None of them works I think. The current common rule doesn't work. They changed their code.

I know that. I'll change the rules a bit.

wout-codes:
What is wrong here?

PR #76000, opened 2021-02-27
Fixes some popups (proxy sites)

wout-codes:
These rules stop random ad-servers from being called. And some ad-servers are blocked here.
1. `https://cinecalidad.proxybit.work/`
2. `videoproxy.icu`
3. `http://hexupload.net/z01v0yoqtdi1`
4. `themovie123.org`


Alex-302 (approved):
If you change something in the middle of the rule, please write about it.

wout-codes:
> If you change something in the middle of the rule, please write about it.

Ok.

Issue #76163, opened 2021-02-28
Add disguised trackers to the Tracking Protection list


ameshkov:
We've finally published the list of actively exploited CNAME-trackers, please examine the trackers in the repo and add those that we can block to the Tracking Protection list:
https://github.com/AdguardTeam/cname-trackers

There are a few important things to note:

1. CNAME trackers should be in a separate file
2. Since the CNAME trackers repo is auto-updated, it'd be better to use `@include` directive in FiltersRegistry
3. It may be not safe to block some of these trackers, please be careful
4. If we need to do additional automatic transformations - please let me know what kind (for instance, batch-adding `$domain=~example.org` exclusions).

PR #76213, opened 2021-03-01
Add CNAME trackers #76163

AdamWr:
https://github.com/AdguardTeam/AdguardFilters/issues/76163

TODO:
- [x] Exclude these trackers which may be unsafe

### DIFF
--- a/SpywareFilter/sections/cname_trackers.txt
+++ b/SpywareFilter/sections/cname_trackers.txt
@@ -1,8 +1,8 @@
 !
 ! CNAME trackers
 !
-! actonservice.com disguised trackers
-!#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/act-on.txt
+! a8.net disguised trackers
+!#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/a8_net.txt
 !
 ! 2o7.net disguised trackers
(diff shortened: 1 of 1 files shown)


Yuki2718:
It seems most of pardot cname tracker are used as click-through so can't entirely be blocked. These are what I confirmed should not be blocked at least as 1st party:
```
! biz-sol.sony.jp
! http://www.hoteresonline.com/articles/6971

! a.nissyoku.co.jp
! https://a.nissyoku.co.jp/l/737513/2019-12-05/pkf2

! biz-info.fielding.co.jp
! https://biz-info.fielding.co.jp/nec-fielding_contact

! biz.mkt.global.dnp.co.jp
! https://biz.mkt.global.dnp.co.jp/l/819623/2020-06-26/2hbbv

! biz.optage.co.jp
! https://biz.optage.co.jp/5g/news/20201012/
```

ameshkov:
I am also not entirely sure about Act-On, I don't see it blocked at all in AG Tracking Protection at the moment (`actonsoftware.com` and `actonservice.com`). Maybe there are other trackers like that as well?

@AdamWr let's start slow and only block those that are 100% blocked in AG Tracking Protection filter, and then we'll slowly add others when we're sure it's okay to block it.

ameshkov (approved):

PR #76413, opened 2021-03-03
add few new trackers, update comments

slavaleleka:


AdamWr:
According to this comment - https://github.com/AdguardTeam/AdguardFilters/pull/76213#issuecomment-789082838
>let's start slow and only block those that are 100% blocked in AG Tracking Protection filter, and then we'll slowly add others when we're sure it's okay to block it.

I didn't add `ebis.ne.jp` and `ad-cloud.jp`, because it's not blocked by `AdGuard Tracking Protection filter`, although `ebis.ne.jp` is blocked by `EasyPrivacy` and `AdGuard DNS filter`, so I guess adding  `ebis.ne.jp` is okay.

ameshkov:
`ebis.ne.jp ` seems to be okay.

Regarding `ad-cloud.jp`, @Yuki2718, is it safe to block it?

Yuki2718:
@ameshkov Should be safe, I have been blocking `||newssuite.sony.net^` but it was used as third-party (`https://git.frogeye.fr/geoffrey/eulaurarien/issues/19`). If you're to add `||ad-cloud.jp^` too, make sure to add `$third-party`. I'll add non-cname ebis tracker to TPF.
Also ref for genieesspv tracker: `https://git.frogeye.fr/geoffrey/eulaurarien/issues/18`

ameshkov:
@slavaleleka I guess we should add `@@||ac.ebis.ne.jp^` to the DNS filter exceptions.

ameshkov (approved):

AdamWr (approved):

PR #76706, opened 2021-03-05
Deleted non-existing websites Dutch Filter

Nomes77:

***Description***:
I deleted websites that doesn't exist anymore or redirects to other sites or for sale. Everything check by hand.

See the sites I deleted, and check if they exist.

[x]: # (This template is meant for missed ad/false positive reports, for other type of reports edit it accordingly)


Alex-302 (approved):

Alex-302:
Hi. You can just write dead domains without pull requests. We will check these domains and remove all rules, or from domains lists if the rule has more than one domain.

PR #76881, opened 2021-03-07
fix #76878 miraculously.ml

### LINKED ISSUE #76878: miraculously.ml
### Issue URL (Anti Adblock Script)
 [https://miraculously.ml/](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fmiraculously.ml%2F)
### Comment
Username: @Milos1171
> I'm getting AdBlock pop-up even when I disable protection on website. I have to turn off AdGuard entirely for it to dissapear.
> --
> The information below is automatically generated. Do not delete it.
> 
> Referrer:
> 
> 
> Custom userscripts:
> https://userscripts.adtidy.org/release/popup-blocker/2.5/popupblocker.user.js
> https://userscripts.adtidy.org/release/assistant/4.3/assistant.user.js
> 

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
Browser: | Opera
AdGuard driver: | WFP
Stealth mode options: | Hide your search queries, <br>Send Do-Not-Track header, <br>Remove X-Client-Data header from HTTP requests, <br>Self-destructing third-party cookies (180), <br>Block WebRTC, <br>Block Push API, <br>Block Location API
DNS filtering: | server: `https://dns.adguard.com/dns-query`<br>
Filters: | <b>Ad Blocking:</b><br/> AdGuard Base<br/><br/><b>Privacy:</b><br/> AdGuard Tracking Protection
Other extensions used: | IDM Integration Module, Tampermonkey, User-Agent Switcher

wout-codes:
#76878


AdamWr (approved):

wout-codes:
@AdamWr Why is this needed?
`miraculously.ml#%#AG_abortInlineScript(/adblock/, 'document.createElement');`

AdamWr:
Comment from - https://github.com/AdguardTeam/AdguardFilters/issues/76878
>I'm getting AdBlock pop-up even when I disable protection on website. I have to turn off AdGuard entirely for it to dissapear.
>DNS filtering: | server: https://dns.adguard.com/dns-query

If user is using AdGuard DNS then redirect rule will not work (at least in this case), because request will be blocked by DNS. So if it's possible then usually I'm trying to add another rule which will fix that problem.

wout-codes:
> Comment from - #76878
> 
> > I'm getting AdBlock pop-up even when I disable protection on website. I have to turn off AdGuard entirely for it to dissapear.
> > DNS filtering: | server: https://dns.adguard.com/dns-query
> 
> If user is using AdGuard DNS then redirect rule will not work (at least in this case), because request will be blocked by DNS. So if it's possible then usually I'm trying to add another rule which will fix that problem.

Okay.

PR #77142, opened 2021-03-10
zn.ua

gork7777:
issue: #77074


Sergey-Lyapin:
@gork7777 
This is not an ad leftover, this element looks the same (empty) without adguard , I've closed the issue.

Issue #75796, opened 2021-02-25
masalaseen.net [NSFW] [Pop-ups]


Nzzyme:
https://masalaseen.net

Pop-ups seen on this domain on Windows and Android. Please check

Nzzyme:
Hi, I am seeing a message "Pop-up Blocked" on Chrome (Android) on almost all the pages and I need to close that box and it comes up again and again which is very annoying. Can you please fix this?

wout-codes:
It's fixed. Please update filters.

Nzzyme:
Thanks. Just observed this empty box on all the pages. Can you please add a rule to remove this?

Example URL: https://masalaseen.net/today-exclusive-sexy-call-girl-blowjob-4/

https://prnt.sc/10kfz2g

wout-codes:
@Nzzyme Can you open the developer tools (CTRL + Shift + l) on the element?

Nzzyme:
I right-clicked on the element and selected "Inspect" and this comes up

https://prnt.sc/10kg8ax

wout-codes:
@Nzzyme Can you try this rule? Check for breakages.
`masalaseen.net##a[href][target="_blank"] > img[width="300"][height="100"]`

Nzzyme:
Works fine with AG & uBlock entension for Chrome. Thanks

Yuki2718:
Fixed by EasyList

Nzzyme:
Can you please check and fix the ads seen on all domains:

Example:
https://masalaseen.net/today-exclusive-desi-boudi-showing-her-boobs-and-pussy/

Screenshot:
https://prnt.sc/11fegfn

PR #77418, opened 2021-03-13
update #75796 masalaseen.net

wout-codes:
#75796


Yuki2718:
You can either remove `[src]` and `[href]` or attribute null string and simplify the path. Since `a` tag almost always has `href` attribute and `img` does `src`, these don't narrow down targets but null string is not so common.

wout-codes:
> 
> 
> You can either remove `[src]` and `[href]` or attribute null string and simplify the path. Since `a` tag almost always has `href` attribute and `img` does `src`, these don't narrow down targets but null string is not so common.

You mean the unknown?

Yuki2718:
`href=""`

wout-codes:
@Yuki2718 `masalaseen.net##center > div > a[href=""][target="_blank"] > img`

Yuki2718 (approved):

PR #77679, opened 2021-03-16
Remove naver.net

ghost:
It is bad for blocking `https://naver.net/wcslog.js` because it breaks a lot of website which use the library.
For example, #77854 was reported because that rule hid a button that a user can purchase an item.


Yuki2718:
Could you point out few more example? I'll open PR for EasyPrivacy.

piquark6046:
Closed because I don't have enough data.

PR #77730, opened 2021-03-17
Fix map.naver.com

ghost:
AdguardTeam/AdguardFilters#77725


SpywareFilter/sections/tracking_servers.txt
```diff
@@ -42,6 +42,7 @@
 ||visilabs.com^$third-party
 ||visilabs.net^$third-party
 !
+||nelo2-col.navercorp.com^$domain=~map.naver.com
```
Sergey-Lyapin, line 45: This rule should be added to tracking-servers_firstparty, I'll move it

PR #77757, opened 2021-03-17
Fixes a popup filmlinks4u.pro

wout-codes:
`https://filmlinks4u.pro/the-girl-on-the-train-2021-movie-online-on-filmlinks4u/`


zloyden (approved):
This is partially fixed. I'll add some rules directly.

PR #77818, opened 2021-03-18
Missed leftover #77708 harpersbazaar.com.tr

wout-codes:
#77708


TurkishFilter/sections/specific.txt
```diff
@@ -42,6 +42,7 @@ yemektarifleri.org,lifebursa.com,mshowto.org,macizlee1.com,macizlee.ne,dunyaatla
 fullhdfilmcibaba.com,720pizlemek.com,1080pfilmizle.net,fullhdfilmcidayi.com,hdfilmizledayi.com,720-pizleme.com,fullhdfilmizlebaba.com,evrenselfilmlerim.net###adStarter
 !
 basarisiralamalari.com##.arkaresimmasas
+harpersbazaar.com.tr###adv728x90web
```
zloyden, line 45: Why not `harpersbazaar.com.tr##.topBannerFrame`?

wout-codes: There may be more leftovers, which I am not sure of. Just because of this missed leftover, that useless reporter, wrote an offensive comment. That's why, I don't wanna hear anymore of those.

zloyden: There is one leftover. Both works on my end, but I'd prefer `harpersbazaar.com.tr##.topBannerFrame`.

zloyden (approved):

PR #77835, opened 2021-03-18
www.giord.ru

gork7777:
issue: #77806


Sergey-Lyapin:
We don't block newsletters in footer, btw too few visitors 4450/month

PR #78037, opened 2021-03-20
fix/update #78032 alibaba.com

wout-codes:
#78032

These three works, so which I one should I pick? I prefer/picked the second one.
```
alibaba.com#?#[campaignid] > div[class] > a[target] > img[crossorigin][src*="?webp=close"]:upward(3)
alibaba.com#?#[campaignid]:has(> div > div[class$="verticalCarousel"])
alibaba.com#?#span[style]:contains(Ad):upward(5)
```

### DIFF
--- a/AnnoyancesFilter/sections/self-promo.txt
+++ b/AnnoyancesFilter/sections/self-promo.txt
@@ -768,7 +768,7 @@ byjus.com###top-banner-register
 byjus.com#?#.sidebar > div.bgc-white:has(> div.relatedlinks-widget > div.widget-content > a[onclick*="sidebar register click"])
 m.naver.com##.id_cui_commemoration_banner
 minhaoperadora.com.br##.minha-conteudo
-alibaba.com#?#[campaignid] > div[class] > a[target] > img[crossorigin][src*="?webp=close"]:upward(3)
+alibaba.com#?#[campaignid]:has(> div > div[class$="verticalCarousel"])
 fotor.com#?##unic-gdpr:upward(2)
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/self-promo.txt
```diff
@@ -768,7 +768,7 @@ byjus.com###top-banner-register
 byjus.com#?#.sidebar > div.bgc-white:has(> div.relatedlinks-widget > div.widget-content > a[onclick*="sidebar register click"])
 m.naver.com##.id_cui_commemoration_banner
 minhaoperadora.com.br##.minha-conteudo
-alibaba.com##a[campaignid][style^="display: block; width: 100%; position: relative; min-height: 120px;"]
+alibaba.com#?#[campaignid]:has(> div > div[class$="verticalCarousel"])
```
Alex-302, line 771: better `alibaba.com##div[campaignid][style*="width: 100%; box-sizing: border-box;"][style*="position: relative; height:"]`
Will look how they chenge banner next time.

Alex-302 (changes requested):

AnnoyancesFilter/sections/self-promo.txt
wout-codes: Will it be long-lasting? Site changes as what the reporter said.

Alex-302: So let's check)

zloyden (approved):

PR #78209, opened 2021-03-22
www.cicekal.net

gork7777:
issue: #78194


TurkishFilter/sections/whitelist.txt
```diff
@@ -1,4 +1,6 @@
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/78194
+cicekal.net#@#.share-container
```
Sergey-Lyapin, line 3: please move to social whitelist

Sergey-Lyapin (changes requested):

gork7777:
This shouldn't be blocked cause this rule is blocking body of site

zloyden (approved):

PR #78701, opened 2021-03-28
fix uBO Assets (#8801) driveddl.xyz

wout-codes:
https://github.com/uBlockOrigin/uAssets/issues/8801


zloyden (changes requested):
Also add `driveddl.xyz##center > div[data-zone][style^="width:"]` or `driveddl.xyz##center > div[class][style="width:320px;height:50px;display: inline-block;margin: 0 auto"]`

zloyden (approved):

PR #79417, opened 2021-04-04
changelly.com

gork7777:
issue: #79390


wout-codes:
@gork7777 This rule doesn't work with AdGuard Apps.

zloyden (approved):

zloyden:
> @gork7777 This rule doesn't work with AdGuard Apps.

The site is in the HTTPS exclusion list.

PR #79447, opened 2021-04-04
www.akbank.com

gork7777:
issue: #79445


zloyden (commented):
Have you checked it with app? It won't help - the site is in the HTTPS exclusion list.

PR #79563, opened 2021-04-05
m.haber7.com

gork7777:
issue: #79562


AdamWr:
Thank you, but it's a mobile website and it seems that it's already fixed in `AdGuard Mobile Ads filter`.

PR #79569, opened 2021-04-05
www.canva.com

gork7777:
issue: #79490


krystian3w:
May no match in:

`https://www.canva.com/pl_pl/edukacja/`

krystian3w (changes requested):
No cover all domain: https://github.com/AdguardTeam/AdguardFilters/pull/79569#issuecomment-813915793

Sergey-Lyapin:
My old similar rule `canva.com##.jU9w5g` still works, so I think we may use `canva.com##._62m9tQ`
I'll add for the rest of cases:
```
canva.com##._1M6VlxSyC3cVhrvvVpX83A
canva.com##._3aI0P4hGOArHwggVa-L1GB
```

PR #79767, opened 2021-04-08
lyrics.az

gork7777:
issue: #79766


EnglishFilter/sections/antiadblock.txt
```diff
@@ -483,7 +483,7 @@ competentedigitale.ro#@#img[width="728"][height="90"]
...
-all3dp.com,keizai.biz,shibukei.com,elmueble.com,comingsoon.it,abc.es,elpais.com,idahostatesman.com,definitions.net,ensonhaber.com,elnacional.com,lapatilla.com,sozcu.com.tr,heraldo.es,20minutos.es,tucson.com,gendai.ismedia.jp,hamakei.com,telva.com,gazettetimes.com,missoulian.com,kenoshanews.com,cumberlink.com,stltoday.com,auburnpub.com,trib.com,dailyprogress.com,journalstar.com,siouxcityjournal.com,pantagraph.com#$#body div.fc-ab-root { display: none !important; }
+lyrics.az,all3dp.com,keizai.biz,shibukei.com,elmueble.com,comingsoon.it,abc.es,elpais.com,idahostatesman.com,definitions.net,ensonhaber.com,elnacional.com,lapatilla.com,sozcu.com.tr,heraldo.es,20minutos.es,tucson.com,gendai.ismedia.jp,hamakei.com,telva.com,gazettetimes.com,missoulian.com,kenoshanews.com,cumberlink.com,stltoday.com,auburnpub.com,trib.com,dailyprogress.com,journalstar.com,siouxcityjournal.com,pantagraph.com#$#body div.fc-ab-root { display: none !important; }
```
zloyden, line 486: This is not necessary, the basic rule is enough.

zloyden (changes requested):

EnglishFilter/sections/antiadblock.txt
zloyden: @gork7777 please take a look at the commit https://github.com/AdguardTeam/AdguardFilters/commit/c7f12514d7638f839cc5f1e1bc34cf2f2a6e1df1 .
This window can be closed, so the rule should be in the AdGuard Annoyances.
In this case, we don't need css rules like above, that's why I chose a basic one.

If the page scrolling was broken, we would add the domain, just like you, but in both rules.

gork7777: Alright I got it, thanks for the info.

zloyden: You are welcome.

gork7777: You're welcome.

Issue #80401, opened 2021-04-14
Add common tracking cookies to the Tracking Protection list


ameshkov:
There're some cookies that are known to be tracking users. It's been a while since we have `$cookie` syntax, but we still don't utilize it.

Note that when AdGuard Tracking Protection list is enabled, most of these cookies can't be set. However, consider the fact that these cookies could've already be set. `$cookie` rules would remove them.

Examples:
* https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage
* Facebook Pixel cookies -- needs to be tested, this may be useful: https://techexpertise.medium.com/facebook-cookies-analysis-e1cf6ffbdf8a
* Doubleclick cookies?
* https://yandex.com/support/metrica/general/cookie-usage.html

Probably, it makes sense to disable AdGuard, visit some websites and look for first-party and third-party cookies set by different scripts.

PR #80418, opened 2021-04-14
Added common tracking cookies #80401

AdamWr:
https://github.com/AdguardTeam/AdguardFilters/issues/80401


wout-codes:
Maybe also add: `$cookie=adEdition` from the following site: `https://www.nbc.com/`.

SpywareFilter/sections/cookies.txt
```diff
@@ -34,4 +33,142 @@
 $cookie=_parsely_visitor
 $cookie=_parsely_session
 ! Tealium Universal tag
-$cookie=utag_main
\ No newline at end of file
+$cookie=utag_main
+! Google - https://developers.google.com/analytics/devguides/collection/analyticsjs/cookie-usage
```
ameshkov, line 37: it would be really cool if we could add a comment to every cookie (where this is possible).

The thing is that those pages you're linking are changing with time and we'll simply lose this info if we don't record it here.

SpywareFilter/sections/cookies.txt
```diff
@@ -34,4 +33,142 @@
...
+$cookie=/^_gat_/
+$cookie=_gat
+$cookie=AMP_TOKEN
+$cookie=/^_gac_/
+$cookie=/^_ga_/
+$cookie=/^_dc_gtm_UA-/
+$cookie=/^CaosGtag_ga/
+$cookie=_gaexp
+$cookie=_gaexp_rc
+$cookie=utma
```
ameshkov, line 51: Should be `__utma` (other `utm*` cookies as well)

ameshkov (changes requested):

Alex-302 (approved):

SpywareFilter/sections/cookies.txt
AdamWr: Right, I will change it.
By the way, this website `https://www.qarson.fr/` is using `utma`, so maybe we should add both rules, a least for `utma`.

image

</details>

AdamWr: Already there is `__utma` and some other rules which starts with `__` below - https://github.com/AdguardTeam/AdguardFilters/pull/80418/files#diff-502809f1020ded4a26ff4ad467ffe53d0e3f42db352794e3537d2910604ea95bR64

SpywareFilter/sections/cookies.txt
Alex-302: Some cookies with description listed in tables:
https://www.staqq.co/cookie-policy
https://support.devex.com/hc/en-us/articles/115000959453-Devex-Privacy-Statement

SpywareFilter/sections/cookies.txt
ameshkov: > Right, I will change it.
> By the way, this website https://www.qarson.fr/ is using utma, so maybe we should add both rules, a least for utma.

Seems to be a custom cookie they're setting.

SpywareFilter/sections/cookies.txt
AdamWr: Added Reblaze and tinypass cookies and some comments.

Alex-302 (approved):

ameshkov (approved):

PR #80891, opened 2021-04-19
Social: inews24.com

ghost:
### URL
http://www.inews24.com/view/1359892

### Screenshot
Screenshot


SocialFilter/sections/specific.txt
```diff
@@ -38,6 +38,9 @@ donbosco-medien.de,giantbomb.com,strefainwestorow.pl,narcity.com,lingohut.com,mt
...
 !
+inews24.com##button[class^="icon svg-facebook"]
```
Sergey-Lyapin, line 41: I think it's better to use onclick attribute in this case:

```
inews24.com##button[onclick="return quick('facebook');"]
inews24.com##button[onclick="return quick('twitter');"]
inews24.com##button[onclick="return quick('band');"]
```

Sergey-Lyapin (changes requested):

PR #81099, opened 2021-04-21
add Branch.io and Fathom to cname-trackers

slavaleleka:
not sure about adding Branch.io because it seems to be used for emails, etc

### DIFF
--- a/SpywareFilter/sections/cname_trackers.txt
+++ b/SpywareFilter/sections/cname_trackers.txt
@@ -16,6 +16,9 @@
 ! AT Internet (formerly XiTi) : at-o.net disguised trackers
 !#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/at-internet-(formerly-xiti).txt
 !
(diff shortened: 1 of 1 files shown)


ameshkov (approved):
@AdguardTeam/filters-maintainers 

I guess that's okay, in line with the policy, but this sparked a good discussion on Twitter: https://twitter.com/ay_meshkov/status/1384886444586643457

Makes me think again - maybe it's time to split the Tracking Protection list into several smaller lists?

* Light: One list for third-party data hogs
* Heavy: One list for third-party semi-essential stuff: crash reports, cookie-less web analytics with no sharing policy (I know a dozen of such), this kind of things
* Heaviest: Self-hosted solutions where nothing is shared with anyone at all.

The point is rather simple - if people will mostly be choosing the "Light" list that would make a point for data hogs - don't be a data hog, for site owners - don't use data hogs, choose better solutions. But again, this will depend on what people will be choosing.

Alex-302 (approved):

PR #81237, opened 2021-04-23
shop.11freunde.de

gork7777:
issue: #81199

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -1221,7 +1221,7 @@ infobus.eu#?#body > #layer1:has(> div#layer2 > div.layer-message-container butto
 portaloswiatowy.pl#$#.overlay-cookie { display: none !important; }
 portaloswiatowy.pl#$#.overlay-hidden-cookie { overflow: visible !important; }
 easycosmetic.de##.footer-cookies
-easycosmetic.de##.overlay
+shop.11freunde.de,easycosmetic.de##.overlay
 easycosmetic.de##.cookies-consent-container
 easybell.de###cookieman-modal
 bramanminiofmiami.com##.consumer-privacy-banner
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -1221,7 +1221,7 @@ infobus.eu#?#body > #layer1:has(> div#layer2 > div.layer-message-container butto
 portaloswiatowy.pl#$#.overlay-cookie { display: none !important; }
 portaloswiatowy.pl#$#.overlay-hidden-cookie { overflow: visible !important; }
 easycosmetic.de##.footer-cookies
-easycosmetic.de##.overlay
+shop.11freunde.de,easycosmetic.de##.overlay
```
Sergey-Lyapin, line 1224: the site is broken with this rule, we need to accept cookies

Sergey-Lyapin (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
wout-codes: Let me help you out @gork7777.

`11freunde.de#%#AG_onLoad(function() {if (!document.cookie.includes("euconsent-v2")) {var g = new MutationObserver(function() {var b = document.querySelector('#czgdpr-notice-wrp button[value="1"]');b&&b.click();});g.observe(document, {childList: !0,subtree: !0});setTimeout(function() {g.disconnect()}, 1E4)}});`

Sergey-Lyapin: why euconsent-v2? 
`#czgdpr-notice-wrp button[value="1"]` this selector matches both buttons

I suggest:
11freunde.de#%#AG_onLoad(function() {if (!document.cookie.includes("gdpr_data")) {var g = new MutationObserver(function() {var b = document.querySelector('#czgdpr-notice-wrp #save');b&&b.click();});g.observe(document, {childList: !0,subtree: !0});setTimeout(function() {g.disconnect()}, 1E4)}});

zloyden (approved):

AnnoyancesFilter/sections/cookies_specific.txt
wout-codes: I made mistake, in my User Rules, it's a different cookie name.

PR #82213, opened 2021-05-03
fix #82161 ndtv.com

### LINKED ISSUE #82161: ndtv.com

***Description***: https://www.ndtv.com/education/aktu-releases-revised-odd-semester-exams-dates has Advertisement leftovers
image

image

wout-codes:
#82161


SpywareFilter/sections/general_url.txt
```diff
@@ -98,6 +98,7 @@ directtrack.*.js
 /watch.*.js?key=*&kw=*&refer=
 /pv_count.
 _analytics.js?
+/static/js/analytics-helper.js
```
Sergey-Lyapin, line 101: I don't see this request, btw it doesn't look popular, so it's better to add to specific with $domain modifier
(or remove this rule if this request does not exist anymore)

Sergey-Lyapin (changes requested):

SpywareFilter/sections/general_url.txt
wout-codes: Let's take risk. The tracking protection filter is so outdated

wout-codes: EasyPrivacy beats it

Sergey-Lyapin: @Alex-302 please decide

Alex-302: https://www.ndtv.com/static/js/analytics-helper.js

Alex-302: But it is just google analytics, low priority issue.

Alex-302 (approved):

PR #82976, opened 2021-05-11
Remove widget.intercom.io

nobelharvards:
widget.intercom.io is a common chat widget used by a lot of websites to facilitate chats between support staff and customers. Sometimes it is the only way of contacting support.

[SpinTel](https://spintel.net.au) is another website that makes use of this widget. It is not practical to have a growing and potentially very large exceptions list for this. The current rule already has 5 exceptions. It might be better to get rid of this rule entirely.

The widget typically appears at the bottom right of the viewport. Sometimes there is a slight delay and only loads after the rest of the page has fully loaded.


chirag127:
> chats between support staff and customers.

Don't Adguard annoyance filter blocks chat pop-up in the corner ?

Alex-302:
This filter is not enabled by default. 
Complaints about blocking widgets are very rare.

PR #83174, opened 2021-05-14
strcloud.in - popups

wout-codes:
`https://strcloud.in/v/A0jRRKRgxgHmxo`


zloyden (changes requested):
@wout-codes also add `strcloud.link`.

EnglishFilter/sections/general_extensions.txt
```diff
@@ -243,6 +243,8 @@ kat.*,kickass.*,kickass2.*,kickasstorrents.*,kat2.*,kattracker.*,thekat.*,thekic
 !
 !
 !
+! strcloud.in - popups
+strcloud.in#%#//scriptlet("prevent-window-open")
```
zloyden, line 247: Do you need this one? This `#%#//scriptlet("abort-current-inline-script", "String.fromCharCode", "/emZnbG9hZGVk|break;case \$\.|break;case \d{1}/")` is enough on my end.

AnnoyancesFilter/sections/tweaks.txt
```diff
@@ -1223,6 +1223,7 @@ procrackerz.org,chapterupdates.com,babelnovel.com,katzen-fieber.de,xn--80aaxitdb
 !+ NOT_PLATFORM(windows, mac, android, ext_ff)
 quotev.com#$#body #res_area ::selection { background-color: #3297FD!important; }
 !--- JS ----
+streamtape.com,strcloud.in#%#//scriptlet("prevent-addEventListener", "contextmenu", "event.dispatch.apply")
```
zloyden, line 1226: What is this rule for?

wout-codes: To fix Anti-right click.

EnglishFilter/sections/general_extensions.txt
wout-codes: I will remove it.

wout-codes: @zloyden 
Btw, the normal HTML Rule for this scriptlet (`#%#//scriptlet("abort-current-inline-script", "String.fromCharCode", "/emZnbG9hZGVk|break;case \$\.|break;case \d{1}/")`) isn't working with the `min-length` and `max-length`. If I remove it and craft it to something like this  `strcloud.*$$script[wildcard="*function*break;case*"]`, then it's working.

zloyden: I have the same. Please check `strcloud.*$$script[wildcard="*function*break;case*"][min-length="30000"][max-length="100000"]`. The script is deleted normally.

AnnoyancesFilter/sections/tweaks.txt
zloyden: Finally reproduced it. Thanks.

EnglishFilter/sections/general_extensions.txt
wout-codes: @zloyden Should I override the current rule or add it under HTML?

zloyden: You can rewrite the rule for the pull request - it's ok here.
Please notice that you should change it in 2 places.
1. Comment section.
2. Actual rule.

wout-codes: Changed.

zloyden (approved):

PR #83577, opened 2021-05-19
Fix/#82614 gamcore.com

stanislav-atr:
#82614


EnglishFilter/sections/specific.txt
```diff
@@ -343,6 +343,8 @@ peoplematters.in##.alertBar
 ||opentrackers.org/i/dediseedbox_sb.png
 ||guidedhacking.com/gh/img/aids/
 ||lowendbox.com/media/banner/
+||getbootstrap.com/docs/*/assets/js/vendor/popper.min.js
+gamcore.com##iframe[src*="loader"]
```
Sergey-Lyapin, line 347: I think we can add:
`||pierlinks.com^$third-party`

Sergey-Lyapin (changes requested):

PR #83787, opened 2021-05-22
Update img*.xyz

wout-codes:
`http://imgfrost.net/aucw5bhn`

Merge it on Friday.


wout-codes:
@Yuki2718 @zloyden 
What do you think of making this rule generic?

`/shrinker/js/h1j.js$script`

Yuki2718:
@wout-codes I already made it generic in uAssets. Since I'm not allowed to merge PR I'll directly add the filter and also a regex added to uAssets, but TBH I don't expect it to last 2 days.

Yuki2718:
Added, you can remove h1j filter from PR.

zloyden:
> What do you think of making this rule generic?

Might be ok, but we should test it. @Yuki2718 if the rule causes issues, you'll rewrite it.

wout-codes:
The rules looks good to me but popups still exist.

zloyden:
> The rules looks good to me but popups still exist.

Try `||kekolangti.com/*/fw.js`.

Yuki2718:
> 
> Try `||kekolangti.com/*/fw.js`.

I don't see that script but they serve different script/frame to different ones. Can you share the full path?

zloyden:
`http://kekolangti.com/fw/fw.js`



> but may need to adjust. How about adding similar regex to Base?

Sure, why not?)

wout-codes:
> > The rules looks good to me but popups still exist.
> 
> Try `||kekolangti.com/*/fw.js`.

@zloyden @Yuki2718 
The simple rule as well as the regex do not work.

wout-codes:
> Try `||kekolangti.com/*/fw.js`.

@Yuki2718 
I actually saw this script in the Filtering Log once. After refreshing the page, it doesn't appear in the Filtering Log anymore.

Yuki2718:
@zloyden @wout-codes Even though doesn't work, I'll add `/\.com\/([a-z]{2})\/\1\.js$/script,~third-party,domain=com` and it's xyz domain version.

zloyden:
@Yuki2718 ok, I saw your commit. What about this pull request?

wout-codes:
This site is trash.

wout-codes:
It has added anti-adblock mechanism.

wout-codes:
@zloyden 
Btw, popups still exist, without scriptlets.

wout-codes:
@Yuki2718 @zloyden 
Blocking this fixes the popups. 
`||pikupalate.com/bb/cc.js`

EnglishFilter/sections/specific.txt
```diff
@@ -1021,8 +1021,6 @@ upload-4ever.com##a[onclick^="window.open("] ~ a[style*="text-align: left;"][sty
 ||banner-api.grouphfm.com/banner/
 ||zoophiliahentai.com/uploads/banners/
 olx.ro,olx.kz,olx.pl,olx.pt,olx.ua,olx.uz,olx.in,olx.com##div[style="min-height: 120px; display: block;"]
-http$third-party,~image,domain=imageking.xyz|imgkoi.xyz|imgair.net|imgko.xyz|imglin.xyz|imgtigr.xyz|imgkr.xyz|imglion.xyz|imgsupa.xyz|imgviv.xyz|imgbird.xyz|imgfsh.xyz|imgtiger.xyz|imgvivo.xyz|imgbrd.xyz|imgfish.xyz|imgsup.xyz|imgkor.xyz|imgfrost.net|imghb.xyz|imgkaida.xyz|imgex.xyz|imgopp.xyz|imgbigo.xyz|imgez.xyz
-/h1j.js$domain=imageking.xyz|imgkoi.xyz|imgair.net|imgko.xyz|imglin.xyz|imgtigr.xyz|imgkr.xyz|imglion.xyz|imgsupa.xyz|imgviv.xyz|imgbird.xyz|imgfsh.xyz|imgtiger.xyz|imgvivo.xyz|imgbrd.xyz|imgfish.xyz|imgsup.xyz|imgkor.xyz|imgfrost.net|imghb.xyz|imgkaida.xyz|imgex.xyz|imgopp.xyz|imgbigo.xyz|imgez.xyz
```
zloyden, line 1025: Why were these rules removed?

zloyden:
> Blocking this fixes the popups.
> `||pikupalate.com/bb/cc.js`

@Yuki2718 added an regex rules to block such requests.

EnglishFilter/sections/specific.txt
wout-codes: There is still a detection. Is the script obfuscated? I seem not to understand it.

wout-codes:
> @Yuki2718 added an regex rules to block such requests.

@zloyden Then explain this:
<details>

image

</details>

zloyden:
> > @Yuki2718 added an regex rules to block such requests.
> 
> @zloyden Then explain this:

I guess those rules should be adjusted. @Yuki2718 please take a look.

Yuki2718:
> Blocking this fixes the popups.
> `||pikupalate.com/bb/cc.js`

That's covered by my regex, though initially had typo.

Yuki2718:
@ameshkov Possible bug in regex engine.

Yuki2718:
And now they mix 4 char pattern - I missed to save the logs but was kinda like `/fghr/fghr.js`>

wout-codes:
So the bug should be fixed.

Yuki2718:
Maybe the current best is to add specific rules, but they change too quickly so they'll soon become obsolete.

wout-codes:
@Yuki2718 Can't the regex be improved?

wout-codes:
@AdamWr 
Could you please help with the Anti-AdBlock? I am not sure what's the site is using :(

AdamWr:
Something like:
`pikupalate.com#@#.ads2`
or
`pikupalate.com#%#//scriptlet('prevent-setTimeout', 'testAd')`
should works.

Yuki2718:
> @Yuki2718 Can't the regex be improved?

@zloyden 's suggestion can block them, but is risky because it also blocks e.g. `http://example.com/ab/cd.js`. I'm not sure about other alternatives.

Yuki2718:
`pikupalate.com#%#//scriptlet('prevent-setTimeout', 'testAd')` is working. @wout-codes Can you add this to PR, with another domain `kekolangti.com`?

zloyden:
> Well, if we specify domain it's safe ofc

That's why we use `$domain` for regex rules.


> I'm not sure about other alternatives.

I added a scriptlet for 2 domains.

EnglishFilter/sections/specific.txt
wout-codes: @zloyden I will add these back due to regex not working.

zloyden: Ok. But it's quite difficult to work with a list when you add a domain in random order, especially when we have a lot of them. I've done this one recently, so let's keep it that way. New domains add left to both lists (url, cosmetics) in the future.

```
pikupalate.com,kekolangti.com,imgflyer.xyz,imgez.xyz,imgbigo.xyz,imgopp.xyz,imgex.xyz,imgkoi.xyz,imageking.xyz,imgweng.xyz,cloudgallery.net,imgair.net,imgkaida.xyz,imgfrost.net,imghb.xyz,imgkor.xyz,imgsup.xyz,imgfish.xyz,imgbrd.xyz,imgvivo.xyz,imgtiger.xyz,imgfsh.xyz,imgbird.xyz,imgviv.xyz,imgsupa.xyz,imglion.xyz,imgkr.xyz,imgair.net,imgtigr.xyz,imglin.xyz
```

zloyden (approved):

zloyden:
> @ameshkov Possible bug in regex engine. Confirmed `/\.com\/([a-z]{2})\/\1\.js$/$script,~third-party` doesn't block `http://kekolangti.com/bb/bb.js` on AG Chrome Extension.

Can you file it to [repo](https://github.com/AdguardTeam/AdguardBrowserExtension)?

wout-codes:
@zloyden It's also a problem in AdGuard Apps. Should not it be reported in the core library repository too?

AdamWr:
Could you please provide example which rule doesn't work and a request which should be blocked?

wout-codes:
As I remember, these rules added did not block a request in the screenshot below. Currently, @Yuki2718 added a tougher regex but only for specific domains.
```
/\.com\/([a-z]{2})\/\1\.js$/$script,~third-party
/\.xyz\/([a-z]{2})\/\1\.js$/$script,~third-party
````

image

AdamWr:
As far as I understand, these rules should block request like this:
`https://pikupalate.com/bb/bb.js`

But this request shouldn't be blocked:
`https://pikupalate.com/bb/cc.js`

https://regex101.com/r/nMvyxe/1

So, if I'm not wrong, it's not a bug.

Thank you for your answer.

zloyden:
> As far as I understand, these rules should block request like this:
> `https://pikupalate.com/bb/bb.js`
> 
> But this request shouldn't be blocked:
> `https://pikupalate.com/bb/cc.js`
> 
> https://regex101.com/r/nMvyxe/1
> 
> So, if I'm not wrong, it's not a bug.
> 
> Thank you for your answer.

I agree with you, but maybe there is another case. That's why I asked for a report.

Yuki2718:
Tested again and `/\.com\/([a-z]{2,4})\/\1\.js$/$script,~third-party` blocked `http://pikupalate.com/hrty/hrty.js` so I guess I just confused bb/bb and bb/cc. @zloyden @AdamWr @wout-codes Please check.

zloyden:
It works as expected on my end (extension / app). If there are no other cases, then don't send a report.

wout-codes:
I don't understand how regex works so I can't help much.

PR #84829, opened 2021-06-03
reddit.com - Add filter for "Download the app" popup

DerrickGibbs1:
Popup is blocking access to the page and cannot be dismissed

***Description***: Block a new popup on reddit.
* **Current behaviour**: reddit has a new popup that is blocking access to the page on mobile browsers.

### DIFF
--- a/AnnoyancesFilter/sections/mobile-app.txt
+++ b/AnnoyancesFilter/sections/mobile-app.txt
@@ -156,6 +156,7 @@ muzofond.fm###sveta__muzika-banner
 m.kenh14.vn##a[href="https://kenh14.vn/download-app.htm"] > img
 moviesjoy.*,himovies.to,tinyzonetv.to###gift-top
 reddit.com##.xPromoAppStoreFooter
+reddit.com##.XPromoBlockingModal
 medium.com##div[id*="ShowPostUnderUser-navbar-open-in-app-button"]
 medium.com##a[href^="https://rsci.app.link"]
 behance.net#$#.js-vue-mobile-app-upsell { display: none!important; }


zloyden:
@DerrickGibbs1 can you please attach a screenshot with this element? Unfortunately, I can't reproduce it.

AnnoyancesFilter/sections/mobile-app.txt
```diff
@@ -156,6 +156,7 @@ muzofond.fm###sveta__muzika-banner
 m.kenh14.vn##a[href="https://kenh14.vn/download-app.htm"] > img
 moviesjoy.*,himovies.to,tinyzonetv.to###gift-top
 reddit.com##.xPromoAppStoreFooter
+reddit.com##.XPromoBlockingModal
```
zloyden, line 159: @DerrickGibbs1 can you please attach a screenshot with this element? Unfortunately, I can't reproduce it.

DerrickGibbs1:
It looks like reddit is showing this ad only on small subs and only on mobile. For example, if you visit https://www.reddit.com/r/videogamephotography/ on an iPhone or with an iPhone user agent it will show up. You may have to reload the page a couple of times. Since posting this filter I have discovered that it does not seem to work with adgaurd for iOS, despite working in uBlock on firefox. There may be a syntax error I do not understand. As you can see this popup blocks the screen and there is no way to dismiss.

screenshot of ad

zloyden:
Thanks for the link, I found it. This works fine on my end, so it can be merged.

zloyden (approved):

PR #85215, opened 2021-06-08
Fix linximpulse.net

fabianski7:

***Description***:
* **Current behaviour**: 

The current rule is not blocking subdomains, and it looks like that domain has several. and they are all a bunch of shit

### DIFF
--- a/SpywareFilter/sections/tracking_servers.txt
+++ b/SpywareFilter/sections/tracking_servers.txt
@@ -138,7 +138,7 @@
 ||pensebig.com.br^$third-party
 ||tracking.m6r.eu^$third-party
 ||analytics.ratopanda.com^$third-party
-||suite*.linximpulse.net^$third-party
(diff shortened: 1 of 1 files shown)


Alex-302:
Yes https://subdomainfinder.c99.nl/scans/2021-06-10/linximpulse.net
But where do you see non-blocked linximpulse.net?

fabianski7:
sometimes while visiting some brazilian sites i find other subdomains. But what is currently blocked is the most common. Looks to me like they alternate between them.

Alex-302:
Better post few sample sites.

fabianski7:
> Better post few sample sites.

On the same site as the issue that reported that site.

As I said, sometimes other domains are used, apparently more like a fallback.

Alex-302:
We cannot just block whole this domain. It is not used for tracking only.

PR #86484, opened 2021-06-22
Fix www.iqiyi.com

DeepChirp:
Test address: https://www.iqiyi.com/v_xyh1a873cw.html
This page is anti-adblock.
image


shenzhiming88:
I tested it on Ublock and the rule works, but I don't know if the rule works on Adguard, if not please change it again.

shenzhiming88:
Fixed in https://github.com/uBlockOrigin/uAssets/commit/6eca7a3dca26ac1f0d94a82ec5f55c3053de36b8

Alex-302:
@shenzhiming88 They added our rules, which are in our Annoyances filter.

PR #86864, opened 2021-06-26
Fix empty space on 01net.com mobile

ryanbr:

***Description***: Blank space on https://www.01net.com/actualites/windows-11-mort-de-mcafee-et-chronologie-des-medias-le-resume-de-l-actu-tech-de-la-semaine-2045147.html mobile UA.
* **Current behaviour**: 

123507376-4d328a00-d669-11eb-9f24-878b72cf13d7

***System configuration***

**Filters:**

`Reported here:`

https://github.com/brave/brave-browser/issues/16629

### DIFF
--- a/FrenchFilter/sections/specific.txt
+++ b/FrenchFilter/sections/specific.txt
@@ -2,6 +2,7 @@
 !-------------------- Specific rules --------------------!
 !--------------------------------------------------------!
 !
+01net.com##.mrf-adv__wrapper
 jeuxvideo.com##.sideWootbox
 actu-environnement.com##.module_annonce
 orange.fr##.autopromo-prospect


FrenchFilter/sections/specific.txt
```diff
@@ -2,6 +2,7 @@
 !-------------------- Specific rules --------------------!
 !--------------------------------------------------------!
 !
+01net.com##.mrf-adv__wrapper
```
zloyden, line 5: I can't find it on the desktop with French IP, could you move the rule to the Mobile filter - specific_web.txt?

zloyden (changes requested):

FrenchFilter/sections/specific.txt
Alex-302: https://github.com/AdguardTeam/AdguardFilters/blob/f21aed9a35bd7d73a53d003886ded77666ff2e9f/MobileFilter/sections/specific_web.txt#L295
I see that element without mobile filter only.
@ryanbr Which filter URL is used in Brave?

ryanbr: we're strictly using the French filterset here

PR #87183, opened 2021-06-30
resolve baidu.com 404

AustinZhang1024:
resolve baidu.com 404 error from search page to target page


Alex-302:
Hi. How to reproduce?

OdinZhang:
> Hi. How to reproduce?

click the search result will return a 404 page of baidu.com instead of the target page

Alex-302:
Better send reports using AdGuard. It seems you haму AdGuard app. It hase some bug, and removeparam rule was disabled yesterday.

PR #87662, opened 2021-07-06
fixes https://github.com/AdguardTeam/AdguardFilters/issues/87638 | do…

ghost:


Alex-302 (changes requested):
hi. Please move domain to the start of the rule(if URL rules with `domain=` - to the end).
Also there are two rules for content blcokers without JS support, below the line
`! for platforms without JS rules support`

IThinkIAmCool:
@Alex-302 done

Alex-302 (approved):

Alex-302:
@IThinkIAmCool Thanks)

PR #88623, opened 2021-07-19
spyhackerz.org

gork7777:
issue: #88574

### DIFF
--- a/TurkishFilter/sections/specific.txt
+++ b/TurkishFilter/sections/specific.txt
@@ -47,7 +47,8 @@ fullhdfilmcibaba.com,720pizlemek.com,1080pfilmizle.net,fullhdfilmcidayi.com,hdfi
 !
 ||paraanaliz.com/wp-content/uploads/*/gedik_gif_banner_youtube.gif
 odatv4.com###menu-icon ~ div[style^="margin-top: "]
-spyhackerz.org##a[href="https://xleet.to/?ref=spy_banner"]
+||i.ibb.co/*/ezgif-com-gif-maker-$domain=spyhackerz.org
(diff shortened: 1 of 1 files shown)


TurkishFilter/sections/specific.txt
```diff
@@ -47,6 +47,7 @@ fullhdfilmcibaba.com,720pizlemek.com,1080pfilmizle.net,fullhdfilmcidayi.com,hdfi
 !
 ||paraanaliz.com/wp-content/uploads/*/gedik_gif_banner_youtube.gif
 odatv4.com###menu-icon ~ div[style^="margin-top: "]
+spyhackerz.org##a[href="https://xleet.to/?ref=spy_banner"]
```
zloyden, line 50: Just `||i.ibb.co/*/ezgif-com-gif-maker-`

zloyden (changes requested):

TurkishFilter/sections/specific.txt
gork7777: What is the difference between them?

zloyden: In this case, it's better to block this gif with size 1.1 Mb.

gork7777: There are two types gif I guess. One is your rule, two is: `i.ibb.co/6HFswqX/900x300.gif`
We can block it like this? `||i.ibb.co/*/900x300.gif`
Second gif is occurs when the page is refreshed several times

zloyden: Yes, also add a modifier `$domain=spyhackerz.org` to those rules.

zloyden (approved):

Issue #88860, opened 2021-07-22
Get rid of the general JS api


ameshkov:
Most of these functions should be replaced with scriptlets:
https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/general_js_api.js

Please check filters and get rid of these:

* [x] `AG_removeElementById`
* [x] `AG_removeElementBySelector`
* [x] `AG_removeParent`
* [x] `AG_each`
* [x] `AG_removeCookie`
* [x] `AG_abortOnPropertyWrite`
* [x] `AG_abortOnPropertyRead`
* [x] `AG_abortInlineScript`
* [x] `AG_setConstant`
* [x] `AG_defineProperty`

If there're any rules that cannot be replaced with a scriptlet, add a comment with an explanation, we'll have to come up with a feasible solution.

AdamWr:
Done.
List of rules which can't be changed to scriplets:
```
! We need a scriptlet to change/set HTML elements attributes, it's part of this task - https://github.com/AdguardTeam/Scriptlets/issues/106
telegraph.co.uk#%#AG_onLoad(function(){AG_each(".gallery.component > .gallery-settings[data-gallery-ad-viewability][data-gallery-slides-per-ad]",function(a){a.setAttribute("data-gallery-ad-viewability","1E9");a.setAttribute("data-gallery-slides-per-ad","1E9")})});

! TODO: replace below rule to scriptlet when this issue will be fixed - https://github.com/AdguardTeam/Scriptlets/issues/57#issuecomment-697240649 (it has been fixed, but apps (stable channel) use old scriptlet library)
konnoznet.xyz#%#AG_abortInlineScript(/adblock/, 'document.createElement');

! Can be replaced by "scriptlet" when this issue will be fixed - https://github.com/AdguardTeam/Scriptlets/issues/57 (it has been fixed, but apps (stable channel) use old scriptlet library)
animex.club#%#AG_abortInlineScript(/document\.getElementById\('babasbmsgx'\)/, 'Math.floor');

! TODO: Change AG_setConstant to scripltet when this issue will be fixed - https://github.com/AdguardTeam/Scriptlets/issues/65
rysafe.blogspot.com#%#AG_setConstant('safelink.counter', '0');
rysafe.blogspot.com#%#AG_setConstant('safelink.click2x', 'false');

! TODO: Change AG_setConstant to scripltet when this issue will be fixed - https://github.com/AdguardTeam/Scriptlets/issues/65
[...]

PR #88880, opened 2021-07-23
Replace AG_removeElementBySelector rules

AdamWr:
https://github.com/AdguardTeam/AdguardFilters/issues/88860

Old rule on `cnet.com` doesn't work, so I have changed it.


zloyden (approved):

zloyden (commented):
Btw, can we use `:remove()` instead of `{ remove: true; }`?

AdamWr:
> Btw, can we use `:remove()` instead of `{ remove: true; }`?

As far as I know, we can use `:remove()`, but if I'm not wrong, it will be converted by filters compiler to `{ remove: true; }`

PR #88935, opened 2021-07-23
vestelinternational.com

gork7777:
issue: #88934


Sergey-Lyapin:
In this case all three rules should be CSS rules (#$#), with your rules scrolling would be broken on iOS.
Rules as they should be:
```
vestelinternational.com#$##cookieModuleModal { display: none!important; }
vestelinternational.com#$#.modal-backdrop { display: none!important; }
vestelinternational.com#$#body { overflow: visible!important; }
```

But anyway this site has too few visitors. (6730/month)

PR #89238, opened 2021-07-26
Made the ad report template less incomprehensible for newcomers to fill out

DandelionSprout:

***Description***:
* **Current behaviour**: 

I've noticed over the past 3 years that people who are new to AdGuard's GitHub repos, are usually rather confused about which lines in this very template to replace, or where (despite the `[//]` tags). So I decided to add better title headers and at least somewhat more clear guidelines.

It would've probably been even better to just as well convert the whole template to the new `.yml` form format, but I wasn't sure how to convert the system config information table in a way that wouldn't make it ≥1500px tall.

image
</details><br/>

* **Expected behaviour**: 

Laymen end-users, and even industry experts, are able to correctly use this repo's GitHub report system. Though to be frank, I presume roughly 75% of them will completely ignore the template anyway, sadly.

image
</details><br/>

N/A

***System configuration***

**Filters:**

N/A

Operating system version (Android/iOS) | 21H1 x64
Browser:                               | Chrome 92.0.4515.107 x64
AdGuard version:                       | N/A
Filters enabled:                       | N/A
Filtering quality (Android only):      | N/A

### DIFF
--- a/.github/ISSUE_TEMPLATE/Ad_report.md
+++ b/.github/ISSUE_TEMPLATE/Ad_report.md
@@ -5,49 +5,49 @@ about: Report an issue with AdGuard Filters
 
 [//]: # (***You can delete or ignore strings starting with "[//]:" They will not be visible either way.)
 
-***Description***:
+———***Description***:———
 * **Current behaviour**: 
 
-[//]: # (Substitute this line with a description of the problem)
+•••Substitute this very line with a description of the problem.•••
 
-[//]: # (Replace %screenshot_url% below with a link to the screenshot of the problem. Also, you can paste image from clipboard instead. It will be automatically loaded.)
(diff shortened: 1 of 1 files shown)


peace2000:
I felt that report template was bit confusing too...

Alex-302:
Current template is ok.

> Though to be frank, I presume roughly 75% of them will completely ignore the template anyway, sadly.

The most reports made using reporting tool, and most of them are correct.

peace2000:
I made a report yesterday by using reporting tool because I felt it was a bit less complicated option than issue report.

Alex-302:
AdGuard sends all necessary info to reporting tool automatically. In general user must select problem type, copypaste URL and screenshot only.

PR #89817, opened 2021-08-01
Move a rule from JPF to TPF/Mobile

ghost:
- `||smartnews-ads.com^$third-party` is essential to block ads in the news app `jp.gocro.smartnews.android`.
- `jp.gocro.smartnews.android` is an outstandingly influential app in Japan, and also popular in the US to some extent. (Source: [AppAnnie JP](https://www.appannie.com/jp/apps/google-play/app/jp.gocro.smartnews.android/), [AppAnnie US](https://www.appannie.com/en/apps/google-play/app/jp.gocro.smartnews.android/))
- `smartnews-ads.com` also appears in Japanese web, but as far as I know, it is for tracking. (Example: https://github.com/easylist/easylist/commit/78c171a)

### DIFF
--- a/MobileFilter/sections/adservers.txt
+++ b/MobileFilter/sections/adservers.txt
@@ -137,6 +137,7 @@
 .ad.kakaocorp.com^
 .ad.kakao.com^
 !
+||smartnews-ads.com^$third-party
 ||ads.iosappsworld.com^
 ||ads-cloud.rovio.com^
 ||mobnatives.com^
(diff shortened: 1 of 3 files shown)


MobileFilter/sections/adservers.txt
```diff
@@ -137,6 +137,7 @@
 .ad.kakaocorp.com^
 .ad.kakao.com^
 !
+||smartnews-ads.com^$third-party
```
zloyden, line 140: If it's popular not only in Japan, you should move it to the Base filter.

zloyden (changes requested):

Yuki2718:
@ryanbr How about adding to EL given the argument above?

zloyden:
Ok. I did it directly. Thanks for the pull request.

PR #89818, opened 2021-08-01
added icon's notification CSS rule ( => dobreprogramy.pl )

awslabspl:
Adds support for removal of icon's notification


AdamWr:
I think that these kind of elements might be useful for some users, so it shouldn't be blocked.
Please add this rule to user filter:
`forum.dobreprogramy.pl###quick-access-notifications > ul > li.read`

awslabspl:
> I think that these kind of elements might be useful for some users, so it shouldn't be blocked.
> Please add this rule to user filter:
> `forum.dobreprogramy.pl###quick-access-notifications > ul > li.read`

Done. Thank you 👍

PR #89890, opened 2021-08-02
de.diesel.com

gork7777:
issue: #89773


BlazDT:
Did not see your pull request. Added domain to existing rule for parent element of your selector.

PR #90069, opened 2021-08-04
vibuonline.de

gork7777:
issue: #90018


AnnoyancesFilter/sections/popups.txt
```diff
@@ -106,6 +106,7 @@ sodapdf.com##.be-tooltip
 !
 ! Specific sites
 !
+vibuonline.de###cookieModalOverlay
```
zloyden, line 109: Wrong section. It should be `cookies.txt`.

zloyden (changes requested):

PR #90885, opened 2021-08-13
Move to correct place with improvement

Yuki2718:
This filter catches a specific cluster of revolving adservers and complements `/ntfc.php?$script` in Base, I don't know why it's in Annoyance. There are some variants in `tag.min.js?z=` so I removed the part and instead limited scope to 3p script.

An example: `https://publicwww.com/websites/%22propu.sh%22/`


Alex-302:
But `/pfe/current/` used by push notifications, also `||propu.sh^` exists in Easylist.

Yuki2718:
If fact 99% of `/pfe/current/` servers are included in EasyList, but the sense of revolving servers is that they change every day and EL can miss some.

Alex-302:
> /pfe/current/

all these are related to propu.sh? If "yes" - better add with a short comment

Yuki2718:
> 
> all these are related to propu.sh? If "yes" - better add with a short comment

Rather hundreds of such domain but yes, okay.

Alex-302 (approved):

PR #91368, opened 2021-08-17
mundo.sputniknews.com

gork7777:
issue: #91343


gork7777:
But this cookie is available on these sites: https://pl.sputniknews.com/
https://it.sputniknews.com/
https://sputniknews.lt/
and etc...
How can we block these?

Sergey-Lyapin:
@gork7777
please change in cookie rule
mundo.sputniknews.com to sputniknews.*

Issue #91340, opened 2021-08-17
https://press.princeton.edu


chirag127:
**Issue URL (Annoyance)**: https://press.princeton.edu/books/hardcover/9780691175577/how-to-die

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/76880977/129684270-f432c312-9c60-4c3c-88f9-4d4c0c359698.png)

</details><br/>

<details><summary>System configuration:</summary>

![image](https://user-images.githubusercontent.com/76880977/126797496-0434116b-fc94-42c9-8139-fba02a985cf2.png)
</details><br/>

PR #91369, opened 2021-08-17
press.princeton.edu

gork7777:
issue: #91340

### DIFF
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -112,6 +112,8 @@ sodapdf.com##.be-tooltip
 !
 ! Specific sites
 !
+press.princeton.edu#$#body { overflow: visible!important; }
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/popups.txt
```diff
@@ -2855,7 +2857,7 @@ hackernoon.com##.js-stickyFooter
 bostonglobe.com#$#body { overflow:visible!important; }
 bostonglobe.com#$#.meter-social-connect { display:none!important; }
 spravochnick.ru#$#body { overflow: visible!important; }
-spravochnick.ru##.m-popup
+press.princeton.edu,spravochnick.ru##.m-popup
```
zloyden, line 2860: All rules will have to be written using css property. if we have issue with page scrolling, otherwise you'll break the site on content blockers. I'll rewrite those rules.

zloyden (changes requested):

zloyden (approved):

PR #91540, opened 2021-08-18
update track parameter filter with: mbid, utm_brand

cornelpv:

***Description***: url tracking parameters _mbid_ and _utm_brand_ not filtered.
* **Current behaviour**: Got this url from facebook feed, and it got those 2 params https://www.wired.co.uk/article/quit-remote-working-dream?mbid=social_facebook&utm_brand=wired

<br/>

* **Expected behaviour**: Should remove _mbid=social_facebook&utm_brand=wired_ from url, result https://www.wired.co.uk/article/quit-remote-working-dream

<br/>


zloyden (approved):

zloyden:
> * Got this url from facebook feed

Generally, it should be specific rule, but you got it from the facebook feed, I suppose there can be other links with the same parameters.

PR #91571, opened 2021-08-19
Blocked zhangxinxu.com homepage ads

sxyazi:
***Description***:
* **Current behaviour**: 

image
</details><br/>

* **Expected behaviour**: 

image
</details><br/>

Visit https://www.zhangxinxu.com u can see

***System configuration***

**Filters:**

Operating system:                      | macOS
Operating system version (Android/iOS) | 11.5.2
Browser:                               | Google Chrome
AdGuard version:                       | 2.5.3.955

### DIFF
--- a/AnnoyancesFilter/sections/antiadblock.txt
+++ b/AnnoyancesFilter/sections/antiadblock.txt
@@ -570,6 +570,7 @@ hgtv.com#@#.showads
 !
 !
 zhangxinxu.com#?#body > *:not(div):not(script):matches-css(width:336px)
+zhangxinxu.com#?#[class^="col-left"] a[rel*="nofollow"], [class^="col-right"] a[rel*="nofollow"]
 miniwebtool.com###blocknotice
 miniwebtool.com#%#//scriptlet("set-constant", "adBlockEnabled", "false")
 deine-tierwelt.de#@#.adunit


AnnoyancesFilter/sections/antiadblock.txt
```diff
@@ -570,6 +570,7 @@ hgtv.com#@#.showads
 !
 !
 zhangxinxu.com#?#body > *:not(div):not(script):matches-css(width:336px)
+zhangxinxu.com#?#[class^="col-left"] a[rel*="nofollow"], [class^="col-right"] a[rel*="nofollow"]
```
zloyden, line 573: We don't add rules with several selectors together, also there is no need for using extended-css `#?#`, wrong section for the rule. It's self-promo.
See how it should be https://github.com/AdguardTeam/AdguardFilters/commit/706a8979619d8ee55e5d317b35f830fb0fd228a7

sxyazi: Okay, but apparently it uses "dynamic class" for the entire page with only ads and forces `display:block !important`, isn't that an anti-adblock? Or is there another reason?

zloyden: > forces `display:block !important`

In this case you can use `#?#` or other variants to avoid this. I exchanged one rule for apps.

> Okay, but apparently it uses "dynamic class" for the entire page with only ads and forces `display:block !important`, isn't that an anti-adblock? Or is there another reason?

You can see those elements without AdGuard, so no. I suppose he uses this property to avoid blocking - mostly, the basic rules `##` don't work against this.

sxyazi: I get it, thanks

PR #91914, opened 2021-08-21
Anti-AdBlock: temporarymail.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/91911


Yuki2718 (changes requested):
We usually add rule(s) for Contents Blocker (where scriptlet is not available) too in the same section, in this case `@@||temporarymail.com/js/prebid-ads.js`.

zloyden (approved):

PR #92372, opened 2021-08-24
hotline.ua

janafrank:
annoying ads in the catalogue


Alex-302:
Hi. I don't see these elements. Please provide sample links.
Also no need to create a section for this site. Just add below the line 53
https://github.com/AdguardTeam/AdguardFilters/blob/master/RussianFilter/sections/specific.txt#L53

If these elements are not third-party ad(self-promo) we add the rules to https://github.com/AdguardTeam/AdguardFilters/blob/master/AnnoyancesFilter/sections/self-promo.txt

janafrank:
Yeah, my fault - I've overlooked the AnnoyancesFilter (thought all regional filters are lumped together).

janafrank:
image

PR #92440, opened 2021-08-25
Remove false positive: a.run.app

ptxmac:
*.run.app is a domain used by google for their cloud run product

***Description***:
* **Current behaviour**: 

Any service hosted on Cloud Run is blocked. [Cloud run](https://cloud.google.com/run) is a service google provides for running microservices. By default a cloud run service gets a domain in the form of `<service-name>-<some-id-google-decide>.run.app`. 

* **Expected behaviour**: 

Micro-service infrastructure provided by google should not be blocked.


ptxmac:
The filter was added in b07fad3d80f3cb2d9e97e3024d00c5234996f32a

Alex-302 (approved):

PR #93384, opened 2021-09-02
Merge antiblock.org exception into single filter

Yuki2718:
It truned out even in the same site there can be several patterns.  There are already some duplicate domain over different extensions (e.g. `touhoudougamatome.blog.fc2.com` is in both jpg and gif filters), but having revisited #92620 `ear-phone-review.com` and I now see gif filter is needed and according to @hirorpt jpg filter is also needed. Given these it'll be better to merge them into single filter, which also reduces our future workload. IIRC iOS CB API supports basic regex.


Alex-302:
Please provide fresh example. Want to check in Safari.

Yuki2718:
@Alex-302 https://github.com/AdguardTeam/AdguardFilters/issues/92620

Alex-302 (approved):

Alex-302:
@Yuki2718 I don't see request, matched by this rule(without AdGuard.)

Yuki2718:
> @Yuki2718 I don't see request, matched by this rule(without AdGuard.)

That's strange, how about `dokuo666.blog98.fc2.com`?

Alex-302:
Reproduced.

Issue #93371, opened 2021-09-02
ac.uk


neer9:
**Issue URL (Annoyance)** :https://www.port.ac.uk/study/postgraduate-research/research-degrees/phd/explore-our-projects/incels-involuntary-celibates

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/82926303/131826854-e825ac31-2c07-46d0-a40b-4d17e9502c74.png)

</details><br/>

<details><summary>System configuration:</summary>

![image](https://user-images.githubusercontent.com/82926303/127352294-46ac9f21-4436-4822-8702-c7d7c09f0563.png)
</details><br/>

PR #93406, opened 2021-09-02
ac.uk

gork7777:
issue: #93371


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -108,6 +108,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+port.ac.uk##.Alert
```
zloyden, line 111: It will be better to use `port.ac.uk##div[data-module="CookieMessage"]`

zloyden (changes requested):

zloyden (approved):

Issue #93830, opened 2021-09-05
webstaurantstore.com


chirag127:
**Issue URL (Annoyance)**:
https://www.webstaurantstore.com/article/397/types-of-coffee-drinks.html

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/76880977/132135400-13260242-7180-4317-b046-8834ac00796a.png)

</details><br/>

<details><summary>System configuration:</summary>

![image](https://user-images.githubusercontent.com/76880977/126797496-0434116b-fc94-42c9-8139-fba02a985cf2.png)
</details><br/>

Mennaruuk:
Remove mailing list nag only:
`webstaurantstore.com##.fsr-email-prompt`

Remove mailing list nag + photo boxes under:
`webstaurantstore.com##.fsr-side.side-bottom`

PR #93877, opened 2021-09-06
Fix webstaurantstore.com (#93830)

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/93830
Gets rid of mailing list element.

image

I was pondering about the photo boxes under the mailing list element. I'm not sure they fit the definition of annoyances. But here's a handy filter for mailing list element + photo boxes under it:

` webstaurantstore.com##.fsr-side.side-bottom`


zloyden (approved):
@Mennaruuk please have a look where the rule should be.

Issue #93728, opened 2021-09-04
raymond.cc


chirag127:
**Issue URL (Ads)** :https://www.raymond.cc/blog/adfender-block-advertisements-without-plugins/

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/76880977/132103447-9864512f-4e05-43c9-a0f3-07f3c5a3cf28.png)

</details><br/>

<details><summary>System configuration:</summary>

![image](https://user-images.githubusercontent.com/76880977/126797496-0434116b-fc94-42c9-8139-fba02a985cf2.png)
</details><br/>

PR #93898, opened 2021-09-06
fix raymond.cc #93728

chirag127:
https://github.com/AdguardTeam/AdguardFilters/issues/93728


zloyden (approved):

zloyden:
It's subscription, so I rewrited the rule and moved it to the right section.

PR #94164, opened 2021-09-09
Cleanup XH filters

Yuki2718:
I think these are obsolete, but it's impossible to check all the pages of all XH proxy sites. I only checked front page, video page, creators page, and search page of `jp.xhamster.com` and `jp.xhamster11.com` - both on PC (Firefox) and mobile (Firefox, only UA changed to mobile).


Yuki2718:
Fixed `hamsterix.cam` in this branch to avoid conflict.

zloyden (approved):

PR #94255, opened 2021-09-10
www.championat.com

gork7777:
issue: #94112


zloyden (commented):
I've answered in the task. Users should install AdGuard Extra for such sites.

PR #94612, opened 2021-09-13
bolix.pl

gork7777:
issue: #94528

I'm not sure `balance` is popup or annoyance.


zloyden:
> I'm not sure `balance` is popup or annoyance.

It's more like selfpromo, but the website has low visitors.

PR #94618, opened 2021-09-13
www.lte-anbieter.info

gork7777:
issue: #94493


AnnoyancesFilter/sections/popups.txt
```diff
@@ -112,6 +112,7 @@ sodapdf.com##.be-tooltip
 !
 ! Specific sites
 !
+lte-anbieter.info##.teaser-wrapper
```
Sergey-Lyapin, line 115: It's better to move to german specific.

Sergey-Lyapin (changes requested):

PR #95122, opened 2021-09-18
Anti-RightClick: tapas.io

ghost:
### URLs
https://tapas.io/episode/2094355
https://tapas.io/episode/1785468


Sergey-Lyapin (changes requested):
@piquark6046
`tapas.io#%#//scriptlet('prevent-addEventListener', 'contextmenu')`
This scriptlet works, it's better to use it instead of js-rule.

PR #95183, opened 2021-09-19
Fix senpaiediciones.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/95172


Sergey-Lyapin:
@piquark6046 
Please don't forget to add domain to HTML-rule.

PR #95205, opened 2021-09-19
Cookie: pcworld.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/95201


AdamWr:
This website requires to accept cookie notice, otherwise some video players not work.
It's already fixed by JS rule, but unfortunately extension for Firefox and Opera can not update/download new JavaScript rules.
Please add this rule to user filter:
```
cmpv2.pcworld.com#%#AG_onLoad(function(){if(window.location.href.includes("consentUUID")){var g=new MutationObserver(function(){var b=document.querySelector('.message-component > button[title^="I Agree"]');b&&(g.disconnect(),b.click())});g.observe(document,{childList:!0,subtree:!0});setTimeout(function(){g.disconnect()},1E4) }});
```

PR #95280, opened 2021-09-20
Fix samuraiscan.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/95275


Sergey-Lyapin:
@piquark6046 please don't forget to add domain to the HTML rule (`$$script[wildcard="*eval*babasbmsgx*"][min-length="5000"][max-length="20000"]` in this case)

PR #95419, opened 2021-09-21
downdetector.it

marysaotomeprog:
https://github.com/AdguardTeam/AdguardFilters/issues/95409


AdamWr (changes requested):
As I recall, this website requires to accept cookie notice, otherwise comments widget is not loaded.
`#_evidon-barrier-wrapper` it's not necessary on my side (it's fixed by another rule) and it does not fix reported issue.
Please add `downdetector.it` to this rule:
https://github.com/AdguardTeam/AdguardFilters/blob/2395c20c30971d091490e7455422956e09071120/AnnoyancesFilter/sections/cookies_specific.txt#L4935

AdamWr:
Due to no response, I have added this rule directly.

PR #95420, opened 2021-09-21
tko.pl

marysaotomeprog:
https://github.com/AdguardTeam/AdguardFilters/issues/95411


AdamWr (changes requested):
It's a Polish website, so please move rule to a correct section - `foreign.txt`
In this case, you can just add domain to this rule:
https://github.com/AdguardTeam/AdguardFilters/blob/16ea80a6660d50090477a2f9016758dff52716e4/EnglishFilter/sections/foreign.txt#L5871

AdamWr:
Due to no response, I have added this rule directly.

PR #95450, opened 2021-09-22
animesaria.com

scripthunter7:
https://github.com/AdguardTeam/AdguardFilters/issues/95426

### DIFF
--- a/EnglishFilter/sections/antiadblock.txt
+++ b/EnglishFilter/sections/antiadblock.txt
@@ -649,7 +649,7 @@ gmailnator.com#%#//scriptlet("abort-current-inline-script", "onload", "adblock")
 ! https://github.com/AdguardTeam/AdguardFilters/issues/93416
 cryptofans.news#%#//scriptlet('prevent-setTimeout', 'swal')
 ! https://github.com/AdguardTeam/AdguardFilters/issues/95426
-animesaria.com#%#//scriptlet('prevent-setTimeout', 'testAd')
+animesaria.com#@#.adsbygoogle
 ! https://github.com/AdguardTeam/AdguardFilters/issues/94208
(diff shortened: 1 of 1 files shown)


EnglishFilter/sections/antiadblock.txt
```diff
@@ -648,6 +648,8 @@ highload.to#%#//scriptlet('prevent-fetch', 'popcash.net')
 gmailnator.com#%#//scriptlet("abort-current-inline-script", "onload", "adblock")
 ! https://github.com/AdguardTeam/AdguardFilters/issues/93416
 cryptofans.news#%#//scriptlet('prevent-setTimeout', 'swal')
+! https://github.com/AdguardTeam/AdguardFilters/issues/95426
+animesaria.com#%#//scriptlet('prevent-setTimeout', 'testAd')
```
Sergey-Lyapin, line 652: Hello @scripthunter7, it's enough and better to use this rule `animesaria.com#@#.adsbygoogle`, because f.e. AdGuard for iOS doesn't support scriptlets.

Sergey-Lyapin (changes requested):

EnglishFilter/sections/antiadblock.txt
scripthunter7: Hello @Sergey-Lyapin! It is "safe" / good idea to add this class to the exceptions?

Sergey-Lyapin: @scripthunter7 
Yes, because the site uses this class for adblock detection only.

Sergey-Lyapin: I changed the rule by myself I hope you don't mind.

PR #95453, opened 2021-09-22
hdfilmcehennemi address change

ghost:
`||hdfilmcehennemi.net/wp-content/uploads/pageskin2.jpg` removed because it's duplicate


zloyden (commented):
I've removed obsolete rules and there are no ads with AdGuard Turkish.

PR #95710, opened 2021-09-25
redundant

ExtremelyCool:


EnglishFilter/sections/specific.txt
```diff
@@ -322,7 +322,7 @@ freetutorials-us.com##.ultra-archive-grid
 allinonedownloadzz.site##.tdi_114
 allinonedownloadzz.site##a[href^="https://vgw35hwr4w6x.com/"]
 3dsportal.net###rightside > .block > .btl
-skidrowcodexgame.com,skidrowcodexreloaded.com##.footer-sticky
```
Sergey-Lyapin, line 325: It seems this rule is still relevant, I added` skidrowcodexreloaded.com#@#.footer-sticky` and the ad is still there:
relevant

krystian3w (changes requested):

krystian3w:
Bug with download EL to AG Base (script generate 23 tries copy rule stages, so failed 115 Times...)?

PR #96232, opened 2021-10-01
add analytics.appsbnet.com.br

fabianski7:
Found on centauro.com.br

ksnip_6

### DIFF
--- a/SpywareFilter/sections/tracking_servers.txt
+++ b/SpywareFilter/sections/tracking_servers.txt
@@ -40,7 +40,7 @@
 !||i.snssdk.com/log/*
 !||i.snssdk.com/slardar/sdk.js
 !
-||analytics.appsbnet.com.br
+||analytics.appsbnet.com.br^
 ||logging.prefixbox.com^
 ||ads-website.ytsservice.com^
 ||carts.guru^$third-party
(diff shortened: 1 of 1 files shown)


SpywareFilter/sections/tracking_servers.txt
```diff
@@ -40,6 +40,7 @@
 !||i.snssdk.com/log/*
 !||i.snssdk.com/slardar/sdk.js
 !
+||analytics.appsbnet.com.br
```
Sergey-Lyapin, line 43: @fabianski7
Please move to adservers_firstparty

Sergey-Lyapin (changes requested):

SpywareFilter/sections/tracking_servers.txt
fabianski7: thanks for the review, but the added domain appears to be collecting browsing data and not displaying ads. It also seems to me that there is no adservers_firstparty in the SpanishFilter folder

Sergey-Lyapin: @fabianski7
Sorry, I got confused. I meant tracking_servers_firstparty.txt in the SpywareFilter =)

PR #96946, opened 2021-10-11
Added few redirect rules

AdamWr:
Fixes https://github.com/AdguardTeam/AdguardForWindows/issues/3844


zloyden (approved):

AdamWr:
These redirect rules which were added, usually fixes more problems than cause, so I think that it's fine to leave them.

>`||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$script,redirect=googlesyndication-adsbygoogle`

Maybe it's a good idea to add it, but I think that we should test this rule for a while before we did it.

Regarding `daj.jp` and `golf.com`, if I'm not wrong, it could be possible to fix these issues rather easily, so I will open a new issue in "scriptlets" repository.

PR #98117, opened 2021-10-26
bnext.com.tw

TobySkarting:
https://github.com/AdguardTeam/AdguardFilters/issues/98113


zloyden (commented):
Hi! That's the wrong file for the rule, but I made changes directly.

PR #98451, opened 2021-10-29
Fix freevpn4you.net

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/98062


zloyden (commented):
Fixed it directly. There was some leftover and an ad,

PR #98564, opened 2021-10-31
Fix finance.rambler.ru

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/98563


zloyden:
Fixed it directly. It was a social element.

PR #98677, opened 2021-11-01
Anti-AdBlock: savvytime.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/98643


zloyden (approved):
We also block request that related to the `admiral`. I did it manually.

PR #99940, opened 2021-11-13
Fix wikirise.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/99929


zloyden (approved):
I've also added rules for content blockers.

PR #100001, opened 2021-11-13
Fix kendam.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/99912


zloyden (approved):
I've also added a rule for content blockers.

PR #100026, opened 2021-11-13
www.uludagsozluk.com

gork7777:
issue: #99480


Alex-302:
image

Added another rule.

PR #100030, opened 2021-11-13
www.tabletowo.pl

gork7777:
issue: #99687


Alex-302:
Already fixed in `Official Polish filters for AdBlock, uBlock Origin & AdGuard`

PR #100035, opened 2021-11-13
trudvsem.ru

gork7777:
issue: #99841


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -110,6 +110,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+trudvsem.ru##.cookies__card
```
zloyden, line 113: I've added a rule directly, please take a look at my commit.

PR #100070, opened 2021-11-14
Fix havoc-os.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/100053


zloyden (approved):
You forgot to add a rule for app. I've done it directly.

PR #100346, opened 2021-11-17
Fix findjob.co.kr

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/100335

### DIFF
--- a/AnnoyancesFilter/sections/self-promo.txt
+++ b/AnnoyancesFilter/sections/self-promo.txt
@@ -50,6 +50,7 @@ disk.yandex.*##.mail360-banner
 !-----------------------------------------------
 !
 !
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/self-promo.txt
```diff
@@ -50,6 +50,8 @@ disk.yandex.*##.mail360-banner
 !-----------------------------------------------
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/100335
+findjob.co.kr##.findall_banner
```
zloyden, line 54: It you have a single rule to the site in this file, you can add it without the task comment, but it's ok.

zloyden (approved):

Issue #100492, opened 2021-11-18
www.iobit.com


chirag127:
### Prerequisites

- [X] This site DOES NOT contains sexually explicit material, otherwise use [NSFW-specific form](./new?template=bug_report_NSFW.yml);
- [X] Filters were updated before reproducing an issue;
- [X] AdGuard product version is up-to-date;
- [X] Browser version is up-to-date;
- [X] If the site or app is broken, disabling AdGuard protection resolves an issue.

### What product do you use?

AdGuard Browser Extension

### AdGuard version

3.6.14

### What type of problem have you encountered?

Annoyances — pop-ups, cookie warnings, etc.

### Which browser(s) do you use?

Chrome

### Which device do you use?

Desktop

### Where is the problem encountered?

`https://www.iobit.com/en/iobit-speed-test.php`

### What filters do you have enabled?

---AdGuard filters---

### Add your comment and screenshots

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/76880977/142468131-afe61573-1c2c-4e59-adf0-8c51a895ff08.png)

</details><br/>

<details><summary>System configuration:</summary>

![image](https://user-images.githubusercontent.com/76880977/141497551-099fafbf-933f-44c7-956a-f88fc1bcb5d5.png)

</details><br/>

### Privacy

- [X] I agree to follow this condition

PR #100510, opened 2021-11-19
www.iobit.com

gork7777:
issue: #100492
`popout` is better cause it'll be block main page gift free popup

### DIFF
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -113,6 +113,7 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !######### Subscription popups add to subscriptions.txt ######### 
 !
+iobit.com##.popout
 protanki.tv##main > div.fixed.z-50
 libertycity.net,libertycity.ru##.create_acc
 banglatech24.com##.extra-cls


AnnoyancesFilter/sections/popups.txt
```diff
@@ -113,6 +113,7 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !######### Subscription popups add to subscriptions.txt ######### 
 !
+iobit.com##.popout
```
zloyden, line 116: Can you reproduce it? I don't see it on my end.

gork7777: Yes. Here: 
image

zloyden: Ok. it looks like an ad. What's the end destination? If it leads to another website, please do next.
1. Move your rule to the `Base - specific`.
2. Remove this `iobit.com###itoppopout`.

gork7777: No it doesn't leads to another website, it leads here: https://www.iobit.com/en/driver-booster.php?from=dbsptest

zloyden (approved):

PR #101198, opened 2021-11-25
Fix prelandappslab.com

ghost:
https://github.com/AdguardTeam/AdguardFilters/issues/101176


zloyden (commented):
It has to be blocked via our malware protection.

PR #101339, opened 2021-11-27
Whitelist

ghost:
multiple infractions will result in a permanent ban. Please report personal attacks, racism, misogyny, or harassment you see or experience.


Alex-302:
Now I am thinking to remove even existing rule. Blocked too much parameters, also required for normal working.

Aztekz:
@Alex-302 I do somewhat agree but these sites are owned and operated by China in other countries domains. They do not respect the privacy policy as they claim to do so. There is no account or data deletion process at all. I feel like we should minimize as much tracking as possible? They also include 3rd party site(s) (Alibaba - CDN's) which has no use even when blocked completely, doesn't seem to break anything for me so far.

It is upto you though how you wanna go through?

Alex-302:
@Aztekz But you added exclusions. I guess something is broken?

> these sites are owned and operated by China

This is not a reason to block too much. Billions of users use Alibaba sites. Many parameters can be used for technical purposes, not for analytics. The site in any case know which pages were visited.

Aztekz:
> @Aztekz But you added exclusions. I guess something is broken?
> 
> > these sites are owned and operated by China
> 
> This is not a reason to block too much. Billions of users use Alibaba sites. Many parameters can be used for technical purposes, not for analytics. The site in any case know which pages were visited.

@Alex-302 The issue wasn't China as much as the different domains (Countries where they are operating) used while not respecting the policy of those countries (Data protection, privacy, transfer etc.). Perhaps my way of writing confused you.

Exclusions were added for 1 specific page where it was not displaying the correct format/categories, but did not break the page.

Also i feel `exclusion` will be better suited word to use instead of `whitelist`?

Alex-302:
I think with exclusions blocking rule is useles. We even do not know the purposes of the most of parameters.

Aztekz:
@Alex-302  Thats the problem with these weird parameters (cannot identify their purpose half the time).

 I think exclusion is necessary, because while browsing the site that are not on landing page **`pages.*`** it still has that **`clickTrackInfo=`**, **`themeID=`** param on some other category/list urls, removing it does not seem to affect anything while also not breaking anything but for that specific **`pages.*`** url.

chirag127:
> @Alex-302 I do somewhat agree but these sites are owned and operated by China in other countries domains

by the way, many users of adguard license are from china, adguard should not break anything for chinese user at any cost.

chirag127:
> Thats the problem with these weird parameters (cannot identify their purpose half the time).

well it's not good idea to block things whose purpose is unknown.

PR #101382, opened 2021-11-27
tieba.baidu.com

DeepChirp:
`https://tieba.baidu.com/f?ie=utf-8&kw=Test`
The script is most likely used to promote.
image


Alex-302:
Do you see on the page something  related to this script?

shenzhiming88:
The blocking does not break the page (at least from what I can see) and the script seems to be related to the promoted game ads

Alex-302:
I mean do you see this gave(or something else) when protection is disabled?

shenzhiming88:
> I mean do you see this game(or something else) when protection is disabled?

Not exactly. Because this page pushes different content every time.

Alex-302:
Ok. better move to annoyances.txt

shenzhiming88:
> Ok. better move to annoyances.txt

@Alex-302 But these games are likely to be associated with ads that are not annoying.

PR #101468, opened 2021-11-28
add regex filter removed from EL

Yuki2718:

***Description***:
* **Current behaviour**: 

Related issue: https://github.com/AdguardTeam/AdguardFilters/issues/101390

The regex is a slight modification of a rule removed from EL for unknown reason (maybe just to reduce the number of regex filters or file size), but I have been observing this to catch those missed by EL/Base. It's already added to uAssets (see https://github.com/uBlockOrigin/uAssets/issues/1835#issuecomment-917437303) and no breakage reported.

* **Expected behaviour**: 

Those revolving cloudfront server should be blocked by regex.

Visit `hakaraw.com` with `||d3a00ifauhjdp.cloudfront.net^$badfilter` as I already added the domain blocking rule.

***System configuration***

**Filters:**

Base, Spyware, and Japanese

Browser:                               | Firefox 94.0.2
AdGuard version:                       | 3.6.16

### DIFF
--- a/EnglishFilter/sections/general_url.txt
+++ b/EnglishFilter/sections/general_url.txt
@@ -40,7 +40,7 @@
 /\/t\/[0-9]{3}\/[0-9]{3}\/a[0-9]{4,9}\.js$/$script
 !
 ! ex. https://d3a00ifauhjdp.cloudfront.net/?afiad=933423
-/^https?:\/\/[0-9a-zA-Z]{10,16}\.cloudfront\.net\/\?[a-z]{3,7}=[0-9]{4,8}$/$script,third-party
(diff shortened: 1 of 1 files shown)


ghost (commented):
This rule works on AdGuard Browser Extension 4.0.64 (Firefox Dev 95.0b12) and AdGuard for Windows (Edge 96.0.1054.34).
However, It seems that there are similar one in AdGuard French.
<details>
<summary>Screenshot</summary>



</details>

Yuki2718:
>  It seems that there are similar one in AdGuard French.

I'm aware of this, but probably won't be excluded unless it causes problems. See https://github.com/AdguardTeam/FiltersRegistry/pull/481

piquark6046:
https://github.com/AdguardTeam/AdguardFilters/pull/101468#issuecomment-981077032
Okay.

Alex-302 (approved):

PR #101857, opened 2021-12-02
www.timeturk.com

gork7777:
issue: #101838


Sergey-Lyapin:
I can't reproduce this ad, do you still see it?

Alex-302:
Same. it seems they removed banner.

PR #101887, opened 2021-12-02
Exchange exception rules to the redirect metrika-yandex

zloyden:
Well, I guess it's time to exchange some old rules using redirects `metrika-yandex`.
@AdguardTeam/filters-maintainers please review it and check the websites.
~~I've found two resources that have to be reported `tv.yandex` and `auth.wi-fi.ru` - will do it.~~
Reported:
https://github.com/AdguardTeam/Scriptlets/issues/164
https://github.com/AdguardTeam/Scriptlets/issues/165

---

So, about the exceptions. If you want to unblock only for platforms that do not support redirect rules, please do
```
@@||mc.yandex.ru/metrika/watch.js$domain=example.org
```
If you want fully unblock it
```
||mc.yandex.ru/metrika/watch.js$script,important,redirect=metrika-yandex-watch,domain=~example.org
@@||mc.yandex.ru/metrika/watch.js$domain=example.org
```
In second case. you should report about the incorrect blocking here https://github.com/AdguardTeam/Scriptlets/issues.

---

Please do not merge this pull request, until it's approved and label `In progress` is removed.

### DIFF
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -19,6 +19,9 @@
 ||tags.tiqcdn.com/utag/*/prod/utag.js$domain=radissonhotels.com|post.ch|usnews.com|yemeksepeti.com|krymr.com
 !
 !
+! https://github.com/AdguardTeam/Scriptlets/issues/163
+||mc.yandex.ru/metrika/tag.js$script,important,redirect=metrika-yandex-tag,domain=~coddyschool.com
+||mc.yandex.ru/metrika/watch.js$script,important,redirect=metrika-yandex-watch,domain=~tv.yandex.*|~auth.wi-fi.ru
 ! TODO: replace `~domain.com` by exclusions after ext4.0 release
 ! https://github.com/AdguardTeam/AdguardForWindows/issues/3973
 ! Probably caused by overriding "dataLayer.push" function
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -19,12 +19,9 @@
 ||tags.tiqcdn.com/utag/*/prod/utag.js$domain=radissonhotels.com|post.ch|usnews.com|yemeksepeti.com|krymr.com
 !
 !
-! TODO: remove a modifier `~domain` when the issues below will be fixed
-! https://github.com/AdguardTeam/Scriptlets/issues/163
-! https://github.com/AdguardTeam/Scriptlets/issues/164
-! https://github.com/AdguardTeam/Scriptlets/issues/165
-||mc.yandex.ru/metrika/tag.js$script,important,redirect=metrika-yandex-tag,domain=~coddyschool.com
-||mc.yandex.ru/metrika/watch.js$script,important,redirect=metrika-yandex-watch,domain=~tv.yandex.*|~auth.wi-fi.ru
+! TODO: check the whitelist file from time to time to remove unnecessary exceptions
+||mc.yandex.ru/metrika/tag.js$script,redirect=metrika-yandex-tag
+||mc.yandex.ru/metrika/watch.js$script,redirect=metrika-yandex-watch
 ! TODO: replace `~domain.com` by exclusions after ext4.0 release
 ! https://github.com/AdguardTeam/AdguardForWindows/issues/3973
 ! Probably caused by overriding "dataLayer.push" function
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -19,7 +19,10 @@
 ||tags.tiqcdn.com/utag/*/prod/utag.js$domain=radissonhotels.com|post.ch|usnews.com|yemeksepeti.com|krymr.com
 !
 !
+! TODO: remove a modifier `~domain` when the issues below will be fixed
 ! https://github.com/AdguardTeam/Scriptlets/issues/163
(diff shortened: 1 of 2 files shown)


SpywareFilter/sections/specific.txt
```diff
@@ -19,6 +19,9 @@
 ||tags.tiqcdn.com/utag/*/prod/utag.js$domain=radissonhotels.com|post.ch|usnews.com|yemeksepeti.com|krymr.com
 !
 !
+! https://github.com/AdguardTeam/Scriptlets/issues/163
```
Alex-302, line 22: also add `TODO` comment with the link to the issue in Scriptlets so as not to forget about to remove `~domain`

zloyden: Ok, done.

SpywareFilter/sections/specific.txt
```diff
@@ -19,6 +19,12 @@
 ||tags.tiqcdn.com/utag/*/prod/utag.js$domain=radissonhotels.com|post.ch|usnews.com|yemeksepeti.com|krymr.com
 !
 !
+! TODO: remove a modifier `~domain` when the issues below will be fixed
+! https://github.com/AdguardTeam/Scriptlets/issues/163
+! https://github.com/AdguardTeam/Scriptlets/issues/164
+! https://github.com/AdguardTeam/Scriptlets/issues/165
+||mc.yandex.ru/metrika/tag.js$script,important,redirect=metrika-yandex-tag,domain=~coddyschool.com
+||mc.yandex.ru/metrika/watch.js$script,important,redirect=metrika-yandex-watch,domain=~tv.yandex.*|~auth.wi-fi.ru
```
AdamWr, line 27: Do we really need an `$important` modifier in these rules?

zloyden: Actually, yes, because we have exceptions. I think we can use hints, but, in general, it will be difficult. Maybe for some cases, but not in common.
In this case, this request will be excluded for `auth.wi-fi.ru` and the redirect rule won't work without `$important`.
```
||mc.yandex.ru/metrika/watch.js$script,redirect=metrika-yandex-watch
@@||mc.yandex.ru/metrika/watch.js$domain=auth.wi-fi.ru
```

AdamWr: But `auth.wi-fi.ru` is excluded from this rule too, so redirect rule will be not applied anyway (I do not mean your example, but the real case here - https://github.com/AdguardTeam/AdguardFilters/pull/101887/files#diff-4229244108ff6695dbabadfef8e0d30d66184c10c4cb116e6d73af18009581a1R27).

Now, if `redirect=metrika-yandex-tag/metrika-yandex-watch` is causing any issue and website is broken when one of these scripts are blocked, you have to add domain in two places, one in a `redirect` rule, second in exception, or you can just add exception with `important` modifier (if I'm not wrong, it's supported only by AdGuard).
We have also redirect rules for `google-analytics`, `google-analytics-ga`, `googletagmanager-gtm`, `googletagservices-gpt` - all these rules are without `important` modifier and as far as I can see, we are adding hints (`!+ PLATFORM(ios, ext_android_cb, ext_safari)`) for these websites, if it's causing problem on platforms which do not support `redirect` modifier.

If you prefer this way, then fine.

zloyden: To be honest, I think that in general it's easier to use such rules without hints.
If you want to exclude a request only for platforms that don't support `redirect` modifier, you just add a domain to the common exception rule and the redirect rule will work as usual on supported platforms. In this case, you don't have to limit the exception rule with adding hint.

AdamWr: You do not need to use hints every time, as you have added a section for exceptions, so for example, you can just add (and remove `important` from redirect rules):
```
! For all platforms
! metrika-yandex - common exceptions for tag.js, watch.js
@@||mc.yandex.ru/metrika/tag.js$domain=example.org
@@||mc.yandex.ru/metrika/watch.js$domain=example.net
!
! For platforms which do not support $redirect rules
!#if (adguard_ext_safari || adguard_app_ios || adguard_ext_android_cb)
@@||mc.yandex.ru/metrika/tag.js$domain=domain1|domain2
@@||mc.yandex.ru/metrika/watch.js$domain=domain3
!#endif
!
```
And you will not need to add `~example.org` and `~example.net` to redirect rules.
But as I said, if you prefer this way, then fine :)

zloyden: I was thinking about adding such a structure, but in general I'd prefer use it without hints.
By the way, that's one of the reasons I decided to make changes via the pull request. We can vote about it together, if you want :)

AdamWr: In general I prefer to not use `important` modifier if it's not really needed.
Now you will have exceptions rules available/visible in filters which will be overridden by `redirect` rule with `important`, and I think it may be a bit misleading, also you will need to add `important` modifier to exception if you will want to test an exception rule in "User rules", but it's just my opinion.

zloyden: Ok, will change it.

zloyden: Done.

AdamWr (approved):

Alex-302 (approved):
better move exclusions group to the top

zloyden:
Ok, I did it.

PR #102159, opened 2021-12-06
www.scotthyoung.com

gork7777:
issue: #101585


EnglishFilter/sections/specific.txt
```diff
@@ -291,6 +291,7 @@ techpowerup.com##.site-headr__logo-wrapper + * a[href*="/images/"] > img
 techpowerup.com##div[class*="bar-item"] a[href*="/images/"] > img
 techpowerup.com#%#//scriptlet('remove-attr', 'wdrwgbethte', 'div')
 !
+scotthyoung.com##.ultra-banner-container
```
Alex-302, line 294: their own self-promo

Alex-302 (changes requested):

EnglishFilter/sections/specific.txt
gork7777: So it should be in annoyances self promo?

Alex-302: In general - yes, but it is not annoying.

PR #102201, opened 2021-12-06
Temporarily disabled 2o7.net disguised trackers

AdamWr:
Temporarily disabled 2o7.net disguised trackers, because it breaks some legit websites.
For example, `google.cd` (`||www.google.cd^`) or `amazon.cn` (`||www.amazon.cn^`) are broken.
Also Fotmob app is broken https://github.com/AdguardTeam/AdguardFilters/issues/102192 due to `||api3.fotmob.com^`


slavaleleka (approved):

ameshkov:
This is super weird. I see that there were some garbage DNS records. Looking into it.

ameshkov:
The issue is actually on our side and it will be fixed soon. Please add it back tomorrow.

ameshkov:
@AdamWr it can be returned back now

PR #102325, opened 2021-12-07
debrid-link.fr

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/102235

### DIFF
--- a/AnnoyancesFilter/sections/cookies_general.txt
+++ b/AnnoyancesFilter/sections/cookies_general.txt
@@ -648,6 +648,7 @@
 ###cookies_box
 ###cookies_container
 ###cookies_info
+###cookies_infos
 ###cookies_message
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_general.txt
```diff
@@ -648,6 +648,7 @@
 ###cookies_box
 ###cookies_container
 ###cookies_info
+###cookies_infos
```
Sergey-Lyapin, line 651: the rule should be domain-specific

Sergey-Lyapin: and this rule leaves a white space at the top
leftover

Sergey-Lyapin (changes requested):

PR #102431, opened 2021-12-08
inmobi.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/101793

Снимок экрана 2021-12-08 в 19 38 16

</details><br/>

### DIFF
--- a/EnglishFilter/sections/whitelist.txt
+++ b/EnglishFilter/sections/whitelist.txt
@@ -120,6 +120,8 @@ youtube.com#$##header.ytd-c4-tabbed-header-renderer { position: absolute !import
 inc42.com#@#.tag-social
 @@||piano.io^$domain=inc42.com
 @@||tinypass.com^$domain=inc42.com
+! https://github.com/AdguardTeam/AdguardFilters/issues/101793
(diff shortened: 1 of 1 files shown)


EnglishFilter/sections/whitelist.txt
```diff
@@ -120,6 +120,8 @@ youtube.com#$##header.ytd-c4-tabbed-header-renderer { position: absolute !import
 inc42.com#@#.tag-social
 @@||piano.io^$domain=inc42.com
 @@||tinypass.com^$domain=inc42.com
+! https://github.com/AdguardTeam/AdguardFilters/issues/101793
+@@||publisher.inmobi.com^$domain=inmobi.com
```
zloyden, line 124: I don't think that we should the exception rule -  it's caused by AdGuard DNS. @AdamWr what do you think?

AdamWr: It's also blocked by `AdGuard Mobile Ads filter`, but as far as I understand, it's advertising company, so I'm not sure if it should be unblocked. By the way, their main website is also broken - `https://www.inmobi.com/`
I guess that it would be better if user would add this rule to "User rules".

PR #102701, opened 2021-12-10
Remove #@#.afs_ads

Yuki2718:
Now that `##.afs_ads` is excluded in FiltersRegistry by regex, we should remove all these.


Alex-302 (approved):

Alex-302:
Removed from another filters.

PR #103065, opened 2021-12-13
tsuiran.jp

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/102983


Yuki2718 (changes requested):
Change `tsuiran.jp###layer_board_area` to `||tsuiran.jp/js/vendor/jquery.layerBoard.js`, better to block the cause than the result in this case as this script is solely for that - not likely to cause false positive.

SKEIDs:
> Change `tsuiran.jp###layer_board_area` to `||tsuiran.jp/js/vendor/jquery.layerBoard.js`, better to block the cause than the result in this case as this script is solely for that - not likely to cause false positive.

ok. i did it.

Yuki2718 (approved):

zloyden (approved):

PR #103154, opened 2021-12-14
novate.ru

gork7777:
issue: #102966


Sergey-Lyapin:
Removing this element makes no difference.

PR #103156, opened 2021-12-14
warrobots.com

gork7777:
issue: #102951


zloyden (approved):
This rule can be unstable. Will also add scriptlet.

PR #103291, opened 2021-12-15
kakakumag.com  

SKEIDs:
website URL : https://kakakumag.com/houseware/?id=17763

image
</details><br/>


AnnoyancesFilter/sections/self-promo.txt
```diff
@@ -214,6 +214,9 @@ analyticsvidhya.com##a[id^="stickyban"]
 testbook.com##.post-body-banner a[href="https://testbook.com/pass"] > img
 ! promotional ads
 kakakumag.com##.side-ad-box
+kakakumag.com###_popIn_recommend
```
Sergey-Lyapin, line 217: It seems all these 3 rules changes nothing

SocialFilter/sections/specific.txt
```diff
@@ -47,6 +47,7 @@ originalnews.nico,archives.gov,sportnews.to,nationalchronicle.in,analogictips.co
 sedaily.com,canshop.jp,koreatimes.co.kr,m.chosun.com,sports.donga.com##.sns_area
 !
 dekiru.net##.module-individual-article-item-share
+kakakumag.com##.article-sns-btm
```
Sergey-Lyapin, line 50: There is the same block at the top - so it's better to use `kakakumag.com##div[class^="article-sns-"]`

Sergey-Lyapin (changes requested):

AnnoyancesFilter/sections/self-promo.txt
Yuki2718: Also note popIn is fully blocked only if TPF is enabled, so this kind of filter should go on Spyware > specific.

Yuki2718: > It seems all these 3 rules changes nothing

The reason is that `.container-side-wrap` has fixed height. I don't want to override this just to remove small placeholder at the bottom.

SKEIDs: ok. I delete these 3 rules

SocialFilter/sections/specific.txt
SKEIDs: ok

Yuki2718 (approved):

AnnoyancesFilter/sections/self-promo.txt
Sergey-Lyapin: @SKEIDs Without tracking protection filter the links in the popin widget lead to site's another articles - that's why we don't remove this widget in adblock filters.

PR #103298, opened 2021-12-15
www.pc-koubou.jp

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103210


SKEIDs:
> @SKEIDs www is not needed

ok

PR #103362, opened 2021-12-16
nlab.itmedia.co.jp 

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103354


Yuki2718:
See https://github.com/AdguardTeam/AdguardFilters/issues/103354#issuecomment-995935720

PR #103417, opened 2021-12-16
aztecapornohd.xxx

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/102419


Yuki2718 (changes requested):
Please add `aztecapornohd.xxx##.lst_ft_bn` to Spanish filters.

Sergey-Lyapin:
И эти тоже в испанский:
```
aztecapornohd.xxx###videoss > section:not([class])
aztecapornohd.xxx##div[id^="dvtop"]
```

Yuki2718:
@Sergey-Lyapin On my end `##div[id^="dvtop"]` doesn't hit, instead `##div[id^="dvsingle"]` hits.

Sergey-Lyapin:
@Yuki2718
It seems they use `.dvcss` for ads only so we can use `aztecapornohd.xxx##.dvcss` (also instead of `##.lst_ft_bn`)

Yuki2718:
> It seems they use `.dvcss` for ads only so we can use `aztecapornohd.xxx##.dvcss` (also instead of `##.lst_ft_bn`)

Yep, you're right.

Sergey-Lyapin:
@jellizaveta 
Please remove `aztecapornohd.xxx##.lst_ft_bn` , `aztecapornohd.xxx##div[id^="dvtop"]` and add `aztecapornohd.xxx##.dvcss`

PR #103419, opened 2021-12-16
young18.net

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/102458


zloyden (changes requested):
Also add these rules.
```
||young18.net/ver.js
young18.net##.stall > div[class][id]
```

zloyden (approved):

PR #103459, opened 2021-12-16
www.hobby-museum.com

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103228


Yuki2718 (changes requested):
Please remove `www.` at the head, it's not needed in this (or rather most of) case.

SKEIDs:
> Please remove `www.` at the head, it's not needed in this (or rather most of) case.

done

Yuki2718 (approved):

PR #103593, opened 2021-12-18
carsensor.net

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103557


Sergey-Lyapin:
`carsensor.net###sns-content-btm` is enough

PR #103737, opened 2021-12-19
qualidea.jp 

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103643

`qualidea.jp##.timeline`
<details><summary>Before</summary>

</details><br/>
<details><summary>After</summary>

</details><br/>


Sergey-Lyapin:
I can't find traffic statistics for this site, which probably means that the site has too few visitors.

Sergey-Lyapin:
If the site has too few visitors, we don't add a new specific rule for such sites, but we may add the domain to an existing rule.

SKEIDs:
> If the site has too few visitors, we don't add a new specific rule for such sites, but we may add the domain to an existing rule.

I see. thank you for tell me.

PR #103807, opened 2021-12-19
greatpretender.jp 

SKEIDs:
https://github.com/AdguardTeam/AdguardFilters/issues/103670


Yuki2718:
The change is okay but no stats.

Yuki2718 (approved):

Sergey-Lyapin:
> The change is okay but no stats.

@Yuki2718
In this case we may add the domain to an existing rule (like here).

PR #103922, opened 2021-12-21
topcam.com.tr

gork7777:
issue: #103626


Sergey-Lyapin (changes requested):
@gork7777 
Please add a rule for the popup too.

gork7777:
@Sergey-Lyapin Are you sure? This is an announcement about the corona vaccine

Sergey-Lyapin:
@gork7777 Ok, let's keep it. Thank you.

PR #103923, opened 2021-12-21
vectorizer.com

gork7777:
issue: #103625


Sergey-Lyapin:
Let's ask @Alex-302  when he gets back from vacation =)

PR #103924, opened 2021-12-21
www.taftcoffee.com

gork7777:
issue: #103614


Sergey-Lyapin (changes requested):
Let's add also the rule for whatsapp chat to popups.txt and social popups.txt

PR #103957, opened 2021-12-21
www.kamuajans.com

gork7777:
issue: #103406


SpywareFilter/sections/specific.txt
```diff
@@ -47,6 +47,7 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|tozlu.com|elleshoes.com|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com
 !
+||tebilisim.com/analytics.js
```
Sergey-Lyapin, line 50: let's use `||cdn.tebilisim.com/analytics.js` because this rule will not match this `https://te-tebilisim-com.cdn.tebilisim.com/analytics.js?v=1.56.2` URL in Safari

Sergey-Lyapin (changes requested):
@gork7777 
And please add `||analytics.tebilisim.com^` to tracking_servers_firstparty

SpywareFilter/sections/specific.txt
gork7777: So what about these issues?
#104098
#104100
#104102
`tebilisim` is there too

Sergey-Lyapin: @gork7777 
Let's add `/_ajax/tebilisim.php?`  for them in general_url

Sergey-Lyapin: @gork7777 to general_url in tracking protection (sorry, I forgot to say)

PR #103914, opened 2021-12-21
itmedia.co.jp

gork7777:
issue: #103090

PR #104093, opened 2021-12-22
japan.cnet.com

spiri-leo:
Issue: #104023

### DIFF
--- a/SocialFilter/sections/specific.txt
+++ b/SocialFilter/sections/specific.txt
@@ -425,6 +425,8 @@ russianemirates.com##.socialink_block
 elesquiu.com##.inner-share-bar
 asia.nikkei.com##.article__share-slot
 freesvg.org##.footer-social-big
+japan.cnet.com##.social-item-bottom
+japan.cnet.com##.leaf-social-inner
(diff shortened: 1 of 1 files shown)


SocialFilter/sections/specific.txt
```diff
@@ -425,6 +425,8 @@ russianemirates.com##.socialink_block
 elesquiu.com##.inner-share-bar
 asia.nikkei.com##.article__share-slot
 freesvg.org##.footer-social-big
+japan.cnet.com##.social-item-bottom
+japan.cnet.com##.leaf-social-inner
```
Yuki2718, line 429: This rule seems to cover both, however, we don't block print button etc. https://github.com/AdguardTeam/AdguardFilters/pull/103914#discussion_r773113029
I think in this case it's probably better to go with a generic filter `##a[href^="https://note.mu/intent/post"]`.

Yuki2718: @zloyden @Sergey-Lyapin

zloyden: @Yuki2718 you were right, I did the changes directly.

PR #104114, opened 2021-12-22
Fix eksperhaber.com

bkrucarci:
Fix for https://github.com/AdguardTeam/AdguardFilters/issues/104105


Sergey-Lyapin:
4800 visitors/month

PR #104156, opened 2021-12-22
kadikoygazetesi.com

gork7777:
issue: #104106


Sergey-Lyapin (changes requested):
This site has too few visitors. Please add the domain to the existing rules:
```
##img[width="300"][height="300"]
##.rek
```

PR #104194, opened 2021-12-22
designcorral.com

gork7777:
issue: #102360


Sergey-Lyapin (changes requested):
@gork7777
Please add the domain to this rule also: `##.code-block`

PR #104326, opened 2021-12-24
asuka-xp.com

SKEIDs:
#104299


SocialFilter/sections/specific.txt
```diff
@@ -4707,7 +4707,7 @@ goal.com##footer[data-sponsorship-slot] > .media + .category + time + i[data-rol
 toplanzi.com###socialShareWidget
 toplanzi.com##img[onclick="fbinvite()"]
 lunar.az##.sende_paylas
-sportnews.to,indir.vip###custom_html-4
+asuka-xp.com,sportnews.to,indir.vip###custom_html-4
```
Yuki2718, line 4710: It seems this rule does nothing.

Yuki2718 (changes requested):

SocialFilter/sections/specific.txt
SKEIDs: delete the space above "カテゴリー".
Is this an unnecessary rule?
@Yuki2718

Yuki2718: Do you see visual difference by this rule?

Yuki2718: Possibly I have to check with AG app than AG extension. On my Chromium + AG Ext. there's no difference.

SKEIDs: I delete this domain from that rule.

PR #104373, opened 2021-12-24
print-gakufu.com

gork7777:
issue: #104343


Yuki2718:
Those banners are self-promo. I guess `##.block-top-banner` in particular should not be blocked even in Annoyances.

Yuki2718 (changes requested):
Remove all the rules in JPF and instead add `print-gakufu.com##.banner` to Annoyances > self-promo.

Yuki2718 (approved):

zloyden (approved):

PR #104491, opened 2021-12-25
geeksforgeeks.org

spiri-leo:
Issue: #104490


EnglishFilter/sections/specific.txt
```diff
@@ -291,6 +291,7 @@ techpowerup.com##.site-headr__logo-wrapper + * a[href*="/images/"] img
 techpowerup.com##div[class*="bar-it"] a[href*="/images/"] img
 techpowerup.com#%#//scriptlet('remove-attr', 'wdrwgbethte', 'div')
 !
+geeksforgeeks.org##.textBasedMannualAds
```
zloyden, line 294: Not reproduced on my end, but it seems we should change the rule a bit.

zloyden (changes requested):

EnglishFilter/sections/specific.txt
zloyden: I'll add the rule directly.

PR #104701, opened 2021-12-26
 guide.line.me

SKEIDs:
#104698


Yuki2718 (commented):
In this case it'll be better to keep `guide.` as `line.me` has many subdomain and this rule may not be very safe.

zloyden (approved):

PR #104804, opened 2021-12-27
www.gundembursa.com

gork7777:
issue: #104690


TurkishFilter/sections/specific.txt
```diff
@@ -50,6 +50,7 @@ kocaelidebugun.com###sagAds
 ||turkiyegazetesi.com.tr/images/pageskin/interstitial/libs/ads-mg*.js
 almanilan.com##div.ai-viewports[data-insertion]
 marketingturkiye.com.tr##.ci-ad
+||gundembursa.com/ad-zones.js
```
Sergey-Lyapin, line 53: Let's make a rule `/ad-zones.js$domain=` and please add domains from the existing rules

Sergey-Lyapin (changes requested):

TurkishFilter/sections/specific.txt
gork7777: `/ad-zones.js` there are three more domains for this rule, should I add them as well?

Sergey-Lyapin: @gork7777 yes

PR #104806, opened 2021-12-27
istiklal.com.tr

gork7777:
issue: #104685

### DIFF
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -47,6 +47,7 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|tozlu.com|elleshoes.com|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com
 !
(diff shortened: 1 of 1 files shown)


SpywareFilter/sections/specific.txt
```diff
@@ -47,6 +47,7 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|tozlu.com|elleshoes.com|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com
 !
+||istiklal.com.tr/hit-increment
```
Sergey-Lyapin, line 50: `.com/hit-increment/` > `/hit-increment/`

Sergey-Lyapin (changes requested):

SpywareFilter/sections/specific.txt
zloyden: Added it directly, but if this rule break something on other sites, please change it to the specific one -> `||istiklal.com.tr/hit-increment/`.

PR #105127, opened 2021-12-30
trtspor.com.tr

gork7777:
issue: #105116

### DIFF
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -47,6 +47,7 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|tozlu.com|elleshoes.com|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com
 !
+||trthaber.com/service/api/stats
 ||paypal.com/webapps/mch/cmd/?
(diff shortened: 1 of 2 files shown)


SpywareFilter/sections/specific.txt
```diff
@@ -47,6 +47,7 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|tozlu.com|elleshoes.com|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com
 !
+||trthaber.com/service/api/stats
```
Sergey-Lyapin, line 50: `||istatistik.trthaber.com^` to tracking servers firstparty please

Sergey-Lyapin (changes requested):

SpywareFilter/sections/specific.txt
Sergey-Lyapin: and remove this rule from specific

gork7777: How do we know it's a first-party tracking server?

Sergey-Lyapin: Because this site sends data to istatistik.trthaber.com and because it's name

Issue #105358, opened 2022-01-02
Dead domains


gork7777:
### Prerequisites

- [X] I checked the documentation and found no answer;
- [X] I checked to make sure that this issue has not already been filed;
- [X] This is not an ad/bug report.

### Problem description

Issue: [Dead Domains in Turkish Filter](https://forum.adguard.com/index.php?threads/dead-domains-in-turkish-filter.46111/)

### Proposed solution

Remove domains from filters

PR #105366, opened 2022-01-02
Dead domains

gork7777:
issue: #105358
Some domains are working. I also deleted some similar domains so don't be confused by the different name


Alex-302:
Better do not make PRs about removing more than one domain in long rules, because some intermediate changes may be lost.
Will check and remove.

PR #105412, opened 2022-01-03
www.swtestacademy.com

gork7777:
issue: #100882


EnglishFilter/sections/specific.txt
```diff
@@ -294,6 +294,7 @@ techpowerup.com#%#//scriptlet('remove-attr', 'zdrwgbethte', 'div')
 facebook.com##a[href="/ads/about/?entry_product=ad_preferences"]
 facebook.com##div[role="region"] + div[role="main"] div[role="article"] div[style="border-radius: max(0px, min(8px, ((100vw - 4px) - 100%) * 9999)) / 8px;"] > div[class]:not([class*=" "])
 porner.tv,pornmonde.com##.sources
+swtestacademy.com###block-6
```
zloyden, line 297: The link is not available, I'll add another rule.

gork7777: @zloyden There's a link, just a typo. You can check this: `www.swtestacademy.com/bypass-login-in-selenium-webdriver/`

zloyden: I see now, but `#block-6` is not stable. If you look at other articles from the website, you'll see that the rule above is not enough.
Added the ext-css rule that should block this leftover.

Issue #100741, opened 2021-11-21
nocourses.com


chirag127:
### Prerequisites

- [X] This site DOES NOT contains sexually explicit material, otherwise use [NSFW-specific form](./new?template=bug_report_NSFW.yml);
- [X] Filters were updated before reproducing an issue;
- [X] AdGuard product version is up-to-date;
- [X] Browser version is up-to-date;
- [X] If the site or app is broken, disabling AdGuard protection resolves an issue.

### What product do you use?

AdGuard Browser Extension

### AdGuard version

3.6.14

### What type of problem have you encountered?

Annoyances — pop-ups, cookie warnings, etc.

### Which browser(s) do you use?

Chrome

### Which device do you use?

Desktop

### Where is the problem encountered?

`https://nocourses.com/t/100-days-of-code-the-complete-python-pro-bootcamp-for-2021-free-download/4897`

### What filters do you have enabled?

---AdGuard filters---

### Add your comment and screenshots

<details><summary>Screenshots:</summary>

![image](https://user-images.githubusercontent.com/76880977/142744771-4dbdedda-f13d-4e27-bea8-426131bd14d7.png)

</details><br/>

### Privacy

- [X] I agree to follow this condition

jellizaveta:
Hi! Hiding this popup may break authorization.

chirag127:
> Hi! Hiding this popup may break authorization.

@jellizaveta Yes you are correct, I am not able to sign up or log in after adding following rules. But I don't need to signup for anything on this site, I would have logged in instead of saying that signup pop up is annoyance if I had need to log in.

`nocourses.com###discourse-modal`
`nocourses.com##.modal-backdrop`

Are these rules good or you have a better alternative? BTW, I would appreciate if you could provide the user rules already made by you in future even if they break small things.

zloyden:
@chirag127 the site has low visitors and please note that such issues have a low priority, but try this
```
nocourses.com#%#//scriptlet('set-constant', 'Discourse.SiteSettings.guest_gate_enabled', 'false')
```

PR #105413, opened 2022-01-03
nocourses.com

gork7777:
issue: #100869


chirag127:
login don't work with these rules.

zloyden (commented):
https://github.com/AdguardTeam/AdguardFilters/issues/100741#issuecomment-1017429788

PR #105414, opened 2022-01-03
www.nicesnippets.com

gork7777:
issue: #100835


Alex-302:
This rule breaks content.

PR #105804, opened 2022-01-05
www.flightconnections.com

spiri-leo:
Issue: #105803


gork7777 (changes requested):
Second is too wide rule, remove these and let's use these:
```
flightconnections.com##div[class^="waldo-"]
flightconnections.com##.parallax-container
```

Alex-302 (approved):

PR #105805, opened 2022-01-05
j-net21.smrj.go.jp

spiri-leo:
Issue: #105778


gork7777 (changes requested):
It's better use `j-net21.smrj.go.jp###esodan-banner`
And this probably should be added on popup filter

Yuki2718 (approved):

Yuki2718:
Hmm, but this seems to be rather a chat widget.

AnnoyancesFilter/sections/popups.txt
```diff
@@ -114,6 +114,7 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !######### Subscription popups add to subscriptions.txt #########
 !
+j-net21.smrj.go.jp###esodan-banner
```
Divine-Frog, line 117: It's more popup than a chat widget.

Divine-Frog (approved):

PR #106201, opened 2022-01-09
m.sport-express.ru

spiri-leo:
Issue: #106200


gork7777:
This should be added on `Mobile > specific_web` probably

RussianFilter/sections/specific.txt
```diff
@@ -51,6 +51,7 @@ dota2.ru,meteo.ua,klops.ru,crazymailing.com,lotzon.pro,novosti-n.org,prokazan.ru
 !
 !
 !
+m.sport-express.ru##.mb_10
```
Sergey-Lyapin, line 54: Too wide rule.
It's better to use `m.sport-express.ru##div[style*="width:325px;"][style*="height: 250px;"]` and please add it to mobile filter > specific_web.txt

Sergey-Lyapin (changes requested):

RussianFilter/sections/specific.txt
spiri-leo: Fixed.

PR #106223, opened 2022-01-09
anime.eiga.com

spiri-leo:
Issue: #106221


Yuki2718:
Sorry I missed your PR and directly fixed (I see no difference by `anime.eiga.com###rectAdBtm` as already covered by `eiga.com###subBtmBar-sticky-wrapper`).

PR #106277, opened 2022-01-10
www.sconto.de

gork7777:
issue: #106230


zloyden (approved):
That was fixed, but you added the domain to the current rule - will do some changes.

PR #107134, opened 2022-01-18
Inspired by #107129

DandelionSprout:

***Description***:
* **Current behaviour**: 

As a result of #107129, I figured it'd be a good idea to pre-emptively fix social media buttons before people report them. So I took some generic hiding rules from https://raw.githubusercontent.com/DandelionSprout/adfilt/master/SocialShareList.txt with minor adjustments, and am submitting them here.

image
</details><br/>

* **Expected behaviour**: 

End-users don't see as much need to report social media buttons, because they're hopefully fixed already.

N/A
</details><br/>

1) Open https://github.com/AdguardTeam/AdguardFilters/issues?q=is%3Aissue+is%3Aopen+label%3A%22T%3A+Social+Widget%22

***System configuration***

**Filters:**

N/A, as the issue report is exclusively for AdGuard Social Media Filter.

Operating system:                      | Windows 11 x64
Operating system version (Android/iOS) | N/A
Browser:                               | Chrome 97.0.4692.71 x64
AdGuard version:                       | uBO 1.40.9b1 + IDCAC 3.3.5
Filters enabled:                       | N/A
Filtering quality (Android only):      | High-quality

### DIFF
--- a/SocialFilter/sections/general_elemhide.txt
+++ b/SocialFilter/sections/general_elemhide.txt
@@ -1075,3 +1075,307 @@ designboom.com,livescience.com,pbtech.co.nz,peoplesworld.org,space.com##.fa-face
 ##.sns_share_area
 ##.sns_twit
 ##.sns_view
+! From «Dandelion Sprout's Annoyances List»
+###add_like_bar
+###batchsdk-ui-alert
+###btn-send-fb
+###contents > yt-share-target-renderer:nth-of-type(1n+2)
+###ctoolbar
+###diaspora-button-container
+###divShare
+###fb-share
+###fb_share
+###inline-share-links
+###likebuttons
+###mashfbar-header
+###new-social-share
+###p-sharing
+###pinterest
+###rail-share-bar
+###share-tool
+###share-unit
+###share_bar
+###ShareEntryPoint
+###social-share
+###ss-floating-bar
+###story-socials
+###tell-a-friend
+###top_share
+##.addthis_button_more
+##.addthis_counter
+##.addtoany_content
+##.addtoany_list
+##.addtoany_share
+##.album-dialog__image-share
+##.ArevicoModal-bg
+##.ArevicoModal
+##.basicsSocialActions-socialIcons
+##.blog-post-single-share
+##.btn--facebook
+##.btn-facebook-like
+##.btn-google-plus
+##.btn-messenger
+##.btn-tweet
+##.btn-twitter
+##.button--share
+##.button__facebook
+##.button__instagram
+##.button__linkedin
+##.button__twitter
+##.c-article__byline-social
+##.c-share-tools
+##.dre-share-band
+##.et_social_icons_container
+##.face-share
+##.facebook-like-box
+##.facebook-messenger-share
+##.facebook-page-plugin
+##.facebook-share
+##.facebook[href*="/sharer/"]
+##.facebook_share
+##.facebook_top
+##.fb-page-plugin
+##.fbbsb-button
+##.figure-share
+##.floating-share-bar
+##.follow_like_module
+##.footerSocial
+##.Gallery-Share
+##.globalSharing
+##.googleplusshare
+##.i-more-share
+##.icon--facebook
+##.icon--twitter
+##.js-share-btn
+##.js-share-pinterest
+##.js-share-stumble
+##.js-share-twitter
+##.jw-sharing-icon
+##.like_facebook
+##.linkedinShare
+##.loader-container.shares
+##.mashsb-box
+##.news-share-link
+##.news_share
+##.post_social
+##.product-sharing
+##.recentsocial
+##.sbtn_main
+##.sd-sharing
+##.share-actions
+##.share-after
+##.share-bar
+##.share-before
+##.share-icon
+##.share-icons
+##.share-it
+##.share-item-plus
+##.share-link.js-require-widget
+##.share-list
+##.share-metas
+##.share-networks
+##.share-nw
+##.share-other
+##.share-social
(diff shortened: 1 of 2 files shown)


ameshkov:
Thank you, the idea is not bad.

Could you please also check if there're any duplicates there?
Some of these rules intersect with existing domain-specific rules.

Here's one example: `joemonster.org###fb-share` is now covered by `#fb-share`.

Also, quick question to @AdguardTeam/filters-maintainers. Why do we keep adding per-domain rules for some elements like `#social-share`?

Talking about this: `
lewdgamer.com,zet.com,clubhunters.ru,multiplayer.it,www-m.cnn.com,edition-m.cnn.com,ampproject.org,drom.ru,haberler.com,daenischesbettenlager.at,daenischesbettenlager.de,jysk.al,jysk.am,jysk.ba,jysk.be,jysk.bg,jysk.by,jysk.ca,jysk.co.id,jysk.co.uk,jysk.com,jysk.com.cy,jysk.com.kw,jysk.com.mt,jysk.ch,jysk.cz,jysk.dk,jysk.ee,jysk.es,jysk.fi,jysk.fr,jysk.ge,jysk.gr,jysk.hr,jysk.hu,jysk.ie,jysk.it,jysk.kz,jysk.lt,jysk.lv,jysk.nl,jysk.md,jysk.me,jysk.mk,jysk.no,jysk.pl,jysk.pt,jysk.ro,jysk.rs,jysk.se,jysk.si,jysk.sk,jysk.ua,jysk.vn,jysk-ks.com,jyskthailand.com,pisiffik.gl,rumfatalagerinn.is,skemman.fo###social-share`

DandelionSprout:
I've at least tried to remove duplicates from `specific.txt` now. The catch is that the tool I use for such (https://abpvn.com/ruleChecker/redundantRuleChecker.html), does not take `$generichide` entries from other lists into the equation.

BlazDT:
When converting big rules to generic rules we have to check if there is an existing generichide rule exception too.

DandelionSprout:
I think I've accomplished that too now, by comparing the specific-entries I had removed, with `$generichide` entries in AdGuard Base and EasyList. Thankfully, only 4 sites were affected by this.

AdamWr:
I'm not sure if all these rules should be removed and it looks like that some of generic rules may cause incorrect blocking (I didn't check all rules).

For example, there is `generichide` rule for `onet.pl`, so specific rule shouldn't be removed. As I recall, the same for `fakt.pl` (`generichide` is in one of the Polish filter lists).

If I'm not wrong, this rule is still needed:
```
einfachbacken.de##.social-share-links > a:not(.social-share-links__link--pdf)
```

Another case:
```
thegamer.com,cbr.com,gamerant.com,screenrant.com##.sharing > .sharing-btn:not(.js-show-coments)
```
As far as I understand, it will be covered by `##.sharing` rule, but it causes that a comment button is hidden here:
```
https://www.cbr.com/the-suicide-squad-peacemaker-james-gunn-defends-rick-flag-death/
```

<hr>

>Why do we keep adding per-domain rules for some elements like `#social-share`?

As far as I know, we are adding generic rule only if rule is used on many websites. I guess, because generic rules may cause incorrect blocking.

DandelionSprout:
Regarding `onet.pl`, I've now finished re-instating entries for sites that had `$generichide` in AdGuard filters *other* than AdGuard Base. That task was not easy to do, enough so that I should take the rest of the night off with food and streaming-cartoons.

Regarding `einfachbacken.de` and `thegamer.com,cbr.com,gamerant.com,screenrant.com##.sharing > .sharing-btn:not(.js-show-coments)` (without having tested the example link just yet), I have in recent months become a passionate proponent of adding `:not(html, body)` or `:not(html, body, article)` to generic `##.` rules in pretty much all major lists; but I wasn't sure if AdGuard was sold on that idea just yet.

Regarding `fakt.pl`, I haven't got that far as to re-instate third-party filters' `$generichide` entries' sites. I've turned on the "Allow edits by maintainers" button just in case.

Alex-302:
Checking in user filter on many sites, and did not find new the necessary rules, which not covered by our. It seeems the most elements exist on few sites only.
But found incorrect blockings. For example, on some sites `##.btn-twitter` blocks links to twitter blogs, similar problem with `##.social-icon` - blocks too much.
Also the most of the rules do not have `"` in selectors, and I don't like a large number of generic rules with `*=`  they lead to  drop in performance on some sites(like with enabled AdBlockID) and may block too much.
Removing of `PLATFORM` hints and making some rules as generic - 100% bad idea.

DandelionSprout:
Well, if you guys want to cut down on the amount of open issues for annoyances and social widgets, you guys simply don't have much other choice than to use generic entries big time.

Alex-302:
We need generic rules, but not hundreds at a time)
Even that recently reported Japanese sites may be broken(blocked too much) by similar rules, created for them. And the most of them are not popular, only one user reports them.

Yuki2718:
TBH I have already made them generic if a rule shall be safe by doing so.

Yuki2718:
> Well, if you guys want to cut down on the amount of open and future issues for annoyances and social widgets, you guys simply don't have much other choice than to use generic entries big time.

We can simply ignore them.

chirag127:
> Well, if you guys want to cut down on the amount of open and future issues for annoyances and social widgets, you guys simply don't have much other choice than to use generic entries big time.

I'm using your list from some time but most of the times it also block button like sign in with facebook and sign in with xyz and sometimes even sign up with email button. but I don't remember the example but saying to avoid this kind of blocking for other adguard used and henceforth creating much more reports.

PR #107556, opened 2022-01-20
www.joinhoney.com

gork7777:
issue: #107449


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+joinhoney.com##div[class^="cookieConsentBanner"]
```
Sergey-Lyapin, line 115: @gork7777 
Let's use` joinhoney.com##div[class^="cookieConsentBannerContainer"]` and please remove 2 obsolete rules

Sergey-Lyapin (changes requested):

PR #107558, opened 2022-01-20
abc.com

gork7777:
issue: #107140


Sergey-Lyapin:
@gork7777 check without AdGuard and with an american IP - this is their own ad which not so annoying so it's better not to touch this element.

Sergey-Lyapin:
@gork7777 Ok, Third-party advertising is often displayed inside this google ads iframe, so I decided to add a rule for this leftover to the base filter.

PR #107701, opened 2022-01-21
Fix icodrops.com

bkrucarci:
https://github.com/AdguardTeam/AdguardFilters/issues/107699


gork7777 (changes requested):
Also there's share title, you can add it here:
`##.share-title`

Sergey-Lyapin:
@bkrucarci
I can't reproduce this ad, do you still see it?

bkrucarci:
> @bkrucarci I can't reproduce this ad, do you still see it?

The website seems clean now, but just for now! These ICO Drops happen between two date. If the sponsored ICO's end, then they have to be removed. But if you close this thread as "not reproduced" after some time another sponsored ICO may happen. I left it to your decision.

Alex-302 (approved):

PR #107893, opened 2022-01-23
スピードテスト.jp

spiri-leo:
Issue: #107889


Yuki2718:
Directly fixed as the page has many aliases and there are also placeholders. Thank you for the PR!

PR #107959, opened 2022-01-23
Drop helpscout rule, fixes #107957

### LINKED ISSUE #107957: developer.helpscout.com
### Prerequisites

- [X] This site DOES NOT contains sexually explicit material, otherwise use [NSFW-specific form](./new?template=bug_report_NSFW.yml);
- [X] Filters were updated before reproducing an issue;
- [X] AdGuard product version is up-to-date;
- [X] Browser version is up-to-date;
- [X] If the site or app is broken, disabling AdGuard protection resolves an issue.

### What product do you use?

Other ad blocker

### AdGuard version

2.1.7.0

### What type of problem have you encountered?

Website or app doesn't work properly

### Which browser(s) do you use?

Firefox

### Which device do you use?

Desktop

### Where is the problem encountered?

https://developer.helpscout.com/

### What filters do you have enabled?

AdGuard Annoyances filter

### What Stealth Mode options do you have enabled?

_No response_

```
![Adblock Plus 2.0]
! Checksum: e9/7vPR/02DEn0Idc/ZGpQ
! Title: AdGuard Annoyances filter
! Description: Blocks irritating elements on web pages including cookie notices, third-party widgets and in-page pop-ups.
! Version: 2.1.7.0
! TimeUpdated: 2022-01-20T20:30:36+00:00
! Expires: 4 days (update frequency)
! Homepage: https://github.com/AdguardTeam/AdGuardFilters
! License: https://github.com/AdguardTeam/AdguardFilters/blob/master/LICENSE
```

Rule:

```
||helpscout.net^$third-party
```

This isn't an annoyance, it's a support tool. You can use it for sending emails/starting live chat sessions to get help on a website.

It's not a feedback tool either a la Hotjar... in some cases it's the _only_ support avenue a website has. Do we really want to block that?

image

### Privacy

- [X] I agree to follow this condition

lol768:
Not a "useless widget" in any sense of the phrase, it's a legitimate bona fide support contact form and/or support chat widget. Not a feedback/ratings tool, not an "assistant". Possibly the only support contact method that some websites have.

See #107957 for template

### DIFF
--- a/AnnoyancesFilter/sections/widgets.txt
+++ b/AnnoyancesFilter/sections/widgets.txt
@@ -232,7 +232,6 @@ sks-auto.ru##.fw-container__step__form__design-btn__body
 ||callkeeper.ru^$third-party
 liquidweb.com##footer + div[class^="FixedButton_"]
 aimersoft.com###aimersoft-chat-robot
-||helpscout.net^$third-party
 ||widget-v*.tidiochat.com^$third-party
 ||salesiq.zoho.in^$third-party
 ||widget.getgist.com^$third-party


Alex-302:
We block widgets like that in Annoyances filter, which is not enabled by default.
Usually, during the year we receive 1-2 complaints about the blocking of these widgets. So most users don't mind.

PR #108068, opened 2022-01-24
fix issue template

Sergey-Lyapin:
How about moving the issue URL in the template up so that it is visible in the popover when hovering over the issue number? That would be handy.


Alex-302:
`What type of problem` also can be moved under `Where is the`.

Alex-302:
`Prerequisites` must be the first in any case.

Sergey-Lyapin:
@Alex-302 in this case the url will not be visible in the popover.
gffgsd

Alex-302:
> in this case the url will not be visible in the popover.

Not a problem. Prerequisites in the middle look weird.

PR #108338, opened 2022-01-27
oko.uk

spiri-leo:
Issue: #108315


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+oko.uk###cmplz-cookiebanner-container
```
Alex-302, line 115: Used on many sites. Will add as generic rule.

PR #108362, opened 2022-01-27
porngameshub.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/106189


zloyden (commented):
Actually, ads have loaded from this domain. Will fix it directly.

PR #108423, opened 2022-01-28
cengage - tracking

shreyasminocha:

***Description***: Cengage makes several tracking and analytics requests.

Reproducing this is gatekept behind having an account and probably even an expensive subscription, so here are some screenshots of the offending requests in dev tools. 

<details><summary>Screenshots:</summary>

activity
profile
resource-interactions
analytics

</details><br/>

### DIFF
--- a/EnglishFilter/sections/specific.txt
+++ b/EnglishFilter/sections/specific.txt
@@ -290,6 +290,11 @@ linkjust.com,updatesmarugujarat.in,mexashare.com,links4u.me,linku.us,skips.link,
 !
 !
 !
+||webassign.net/services/cubby/bff/analytics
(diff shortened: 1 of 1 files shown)


Alex-302:
Which site?

Alex-302:
I don't see these requests.

shreyasminocha:
> Reproducing this is gatekept behind having an account and probably even an expensive subscription

It's probably because they're made only when a user is signed in. Here's a video demo:

https://user-images.githubusercontent.com/11537232/152407486-151b6fa3-c118-4f7d-a628-a6eb926b1163.mp4

Alex-302 (changes requested):
Wrong file. Please move to Tracking Protection filter

PR #108629, opened 2022-01-29
karriere.rewe-group.com

spiri-leo:
Issue: #108401


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+karriere.rewe-group.com##.ukCookiePolicy
```
zloyden, line 115: @spiri-leo not reproduced on my end. Is it still relevant?

spiri-leo: It doesn't reproduced for me either.

PR #108632, opened 2022-01-29
www.yedas.com

spiri-leo:
Issue: #108095


urvbcn88:
Do not add rules for sites that have no visitors.
https://pr-cy.io/yedas.com/
https://website.informer.com/yedas.com

PR #108636, opened 2022-01-29
www.derya.com

spiri-leo:
Issue: #108634


urvbcn88:
Do not add rules for sites that have no visitors.
https://pr-cy.io/derya.com/
https://website.informer.com/derya.com
https://www.similarweb.com/website/derya.com/#overview

PR #108714, opened 2022-01-30
nutriu.en.aptoide.com

spiri-leo:
Issue: #108713


Alex-302:
Added `aptoide.com##div[class*="__AdsContainer-"]`, it covers also mobile version.

PR #108753, opened 2022-01-30
Fix www.avast.com

bkrucarci:
https://github.com/AdguardTeam/AdguardFilters/issues/108684

### DIFF
--- a/EnglishFilter/sections/specific.txt
+++ b/EnglishFilter/sections/specific.txt
@@ -292,7 +292,7 @@ linkjust.com,updatesmarugujarat.in,mexashare.com,links4u.me,linku.us,skips.link,
 !
 !
 !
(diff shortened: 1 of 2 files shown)


EnglishFilter/sections/specific.txt
```diff
@@ -292,7 +292,7 @@ linkjust.com,updatesmarugujarat.in,mexashare.com,links4u.me,linku.us,skips.link,
 !
 !
 !
-avast.com###offer-chrome
+avg.com,avast.com###offer-chrome
```
Alex-302, line 295: Not an ad. Better to annoyances.

zloyden: Moved it here.

zloyden (approved):

PR #109061, opened 2022-02-02
www.f-secure.com

spiri-leo:
Issue: #109022


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+||assets.f-secure.com/fs-cookie-consent.js
```
zloyden, line 115: This site is on the https exclusion list. This won't work for apps.
However, we can replace the old rule.

zloyden (approved):

PR #109255, opened 2022-02-04
Fix  m.newspic.co.kr

piquark6046:
https://github.com/AdguardTeam/AdguardFilters/issues/109247


SocialFilter/sections/specific.txt
```diff
@@ -51,6 +51,7 @@ foodallergy.org,sharecuts.cn,wtop.com,thegrammarexchange.infopop.cc,open.online,
...
 !
+m.newspic.co.kr##.share_view
```
zloyden, line 54: This rule is ok, but don't forget about other elements. I've removed leftovers.

zloyden (approved):

SocialFilter/sections/specific.txt
piquark6046: Okay. I will push a patch that resolve the leftover.

PR #109506, opened 2022-02-05
www.haberturk.com

gork7777:
issue: #109419


Alex-302:
Cannot reproduce both elements.

PR #109573, opened 2022-02-06
www.brogle.de

spiri-leo:
Issue: #109558


zloyden (commented):
These rules have to be written via `#$#` syntax else the website breaks on content blockers - fixed directly.

PR #109579, opened 2022-02-06
www.steiner-juwelier.com

spiri-leo:
Issue: #109563


AnnoyancesFilter/sections/subscriptions.txt
```diff
@@ -31,6 +31,8 @@ turkuazgazetesi.net,yenialanya.com,mansetaydin.com,seskocaeli.com,kocaelibarisga
 ##.widget_blog_subscription
 ~mega-debrid.eu,~curvegames.com,~news.un.org,~norwichzen.org.uk,~dereferer.me###mc_embed_signup
 !
+steiner-juwelier.com#$#body { overflow: auto !important }
+steiner-juwelier.com##.remodal-overlay
```
zloyden, line 35: These rules have to be written via `#$#` syntax.
Also there is no need to apply `overflow` to the `body` - fixed directly.

PR #109585, opened 2022-02-06
www.watch.de

spiri-leo:
Issue: #109577

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+watch.de###cookie-notice-block
 veepn.com##.cookie_privacy
 cenyvaptekah.ru##.alert-first-usage
 vogue.com.cn###user_cookie_pop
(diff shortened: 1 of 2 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -112,6 +112,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+watch.de###cookie-notice-block
```
zloyden, line 115: Not reproduced on my end. It's already hidden.

AnnoyancesFilter/sections/subscriptions.txt
```diff
@@ -31,6 +31,8 @@ turkuazgazetesi.net,yenialanya.com,mansetaydin.com,seskocaeli.com,kocaelibarisga
 ##.widget_blog_subscription
 ~mega-debrid.eu,~curvegames.com,~news.un.org,~norwichzen.org.uk,~dereferer.me###mc_embed_signup
 !
+watch.de#$#body { overflow: auto !important }
+watch.de###newspopup_up_bg_9
```
zloyden, line 35: These rules have to be written via `#$#` syntax - fixed directly.

PR #109600, opened 2022-02-06
www.engelhorn.de

spiri-leo:
Issue: #109597

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -112,6 +112,9 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+engelhorn.de#$#body { overflow: auto !important }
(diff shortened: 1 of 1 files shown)


gork7777 (changes requested):
Remove `cmpbox` rules and add domain to here:
`##div[id^="cmpbox"]`

spiri-leo:
> Remove `cmpbox` rules and add domain to here: `##div[id^="cmpbox"]`

Fixed.

gork7777:
Hm it's better to add to existing rule: `kicker.de##div[id^="cmpbox"]`

spiri-leo:
> Hm it's better to add to existing rule: `kicker.de##div[id^="cmpbox"]'

Fixed

zloyden (commented):
@gork7777 @spiri-leo please take a look at my commit https://github.com/AdguardTeam/AdguardFilters/commit/01c477df187fa328d69524a835b72e04cf7e39d8 - this is how it has to be.
There had also been some issue with store, fixed directly.

PR #109885, opened 2022-02-09
aeromexico.com

gork7777:
issue: #109864

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -119,6 +119,8 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+aeromexico.com#$#html { overflow: auto !important; }
+aeromexico.com#$#.Modal { display: none !important; }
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -119,6 +119,8 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+aeromexico.com#$#html { overflow: auto !important; }
+aeromexico.com#$#.Modal { display: none !important; }
```
Sergey-Lyapin, line 123: @gork7777 try to login with this rules

Sergey-Lyapin (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
gork7777: I got the problem but how do we separate them from each other?

zloyden: There you should use a scriptlet - fixed it directly https://github.com/AdguardTeam/AdguardFilters/commit/ab535fd1b2cc223f99fe942c15614a147d97e155.

PR #110009, opened 2022-02-10
mediamarkt.com.tr

gork7777:
issue: #109990
Removed old rules and I added new.

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -119,6 +119,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+mediamarkt.*###mms-consent-portal-container
 mea.com.lb###divAllowCookies
 hoverwatch.com###acceptbar
 first-class-and-more.de###fade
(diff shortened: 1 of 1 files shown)


Alex-302 (approved):

AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -119,6 +119,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+mediamarkt.*###mms-consent-portal-container
```
Alex-302, line 122: Cannot find `#mms-consent-portal-container`

Alex-302 (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
gork7777: @Alex-302 You can find it in here: `www.mediamarkt.at` and also you can find `##.gdpr-cookie-layer--show` rule in `www.mediamarkt.nl`

Alex-302: So `*` is not required here, if used on one domain only.

Alex-302 (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
gork7777: `###mms-consent-portal-container` rule used in `mediamarkt.at` and `mediamarkt.de` sites.

PR #110153, opened 2022-02-11
www.generacodice.com

spiri-leo:
Issue: #110151


AnnoyancesFilter/sections/popups.txt
```diff
@@ -119,6 +119,8 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !######### Subscription popups add to subscriptions.txt #########
 !
+generacodice.com#$#body { overflow: auto !important; }
+generacodice.com#$#.modal-backdrop { display: none !important; }
```
zloyden, line 123: You should add one rule, fixed directly.

PR #110400, opened 2022-02-14
myflixer.id #110170

Fisherman-Ulysses:
#110170


Alex-302:
Added different rule.

PR #110434, opened 2022-02-14
www.dashlane.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/108721

In issue system configuration indicated that user has AdGuard desktop application, and it means that the cosmetic rules will not work on this website (website in https exclusion list). On my mind rule for this website might be helpful
in AdGuard browser extension.

Снимок экрана 2022-02-14 в 16 50 55

</details><br/>


zloyden (approved):
You should also answer about this in the task.

PR #110442, opened 2022-02-14
mangatone.com

gork7777:
issue: #110214


EnglishFilter/sections/specific.txt
```diff
@@ -294,6 +294,7 @@ linkjust.com,updatesmarugujarat.in,mexashare.com,links4u.me,linku.us,skips.link,
 !
 !
 !
+mangatone.com##.c-sidebar
```
zloyden, line 297: I fixed this directly - it's better to use ext-css rules else your rule will block content.

PR #110468, opened 2022-02-14
analyticsindiamag.com

gork7777:
issue: #102924


EnglishFilter/sections/specific.txt
```diff
@@ -294,6 +294,7 @@ linkjust.com,updatesmarugujarat.in,mexashare.com,links4u.me,linku.us,skips.link,
 !
 !
 !
+analyticsindiamag.com##.elementor-widget[data-id="2deba64"]
```
zloyden, line 297: Better use ext-css here, will fix it directly.

PR #110563, opened 2022-02-15
www.koton.com

gork7777:
issue: #110562


SocialFilter/sections/specific.txt
```diff
@@ -53,6 +53,7 @@ foodallergy.org,sharecuts.cn,wtop.com,thegrammarexchange.infopop.cc,open.online,
...
 !
+koton.com##.detailPageTexts:not([class*="productdetail"]):not([class*="fancybox"])
```
Sergey-Lyapin, line 56: I like this rule more:
`koton.com##img.share + a.detailPageTexts`

Sergey-Lyapin (changes requested):

PR #110959, opened 2022-02-19
www.newsit.gr

gork7777:
issue: #110879


AdamWr (approved):

AdamWr:
Cookie notice and push notification seems to be already blocked by Annoyances filter, so I have removed these rules.

PR #110973, opened 2022-02-19
www.sporx.com

gork7777:
issue: #110823


Yuki2718 (commented):
Are these ads? Though I see @Alex-302 blocked the same one, they look like widgets (annoyance). Anyway you can simply adjust the existing rule `||iddaa.sporx.com/web/kutu_iddaa_bulten.php` to `||iddaa.sporx.com/web/kutu_$subdocument` instead of adding new rules.

gork7777:
@Yuki2718 Yes these are ads and big 15 page slider in the middle also has same ad after page 4 (it should slide automatically). 
This rule ok for block that ad? `sporx.com##img[src="https://cdn.sporx.com/img/54/2022/e370c79582a259483a3e683ccdf402f1.jpg"]`

image

Yuki2718:
> big 15 page slider in the middle also has same ad after page 4 (it should slide automatically).

Sorry where is this?

gork7777:
I showed in the picture, the slider should automatically slide side for the reproduce

Yuki2718:
Well, I don't see that element in the issue URL or top page even with Turkish VPN. Can you share the URL of the page, or take a screenshot of devtools around that?

@Alex-302 Do you see https://github.com/AdguardTeam/AdguardFilters/pull/110973#issuecomment-1046230070 ?

PR #111243, opened 2022-02-23
warthunder.com cookie notice (Annoyance)

SeriousHoax:

***Description***:
* **Current behaviour**: 
Cookie Notice

image
)
</details><br/>

* **Expected behaviour**: 
No cookie notice

Visit: `https://warthunder.com/en`

***System configuration***

**Filters:**

Base, Tracking, Annoyance, Social

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -452,6 +452,7 @@ fotostrana.ru##.footer-cookie-agreement
 pokemonmasters-game.com##.c-agree-cookie
 tiktok.com##div[class*="DivCookieBannerContainer"]
 /_js/consent_check.php$domain=sporx.com|superfb.com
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -452,6 +452,7 @@ fotostrana.ru##.footer-cookie-agreement
 pokemonmasters-game.com##.c-agree-cookie
 tiktok.com##div[class*="DivCookieBannerContainer"]
 /_js/consent_check.php$domain=sporx.com|superfb.com
+warthunder.com##.wt-cb
 warthunder.ru##.wt-cb
```
zloyden, line 456: You can just do one rule - will show you here.

zloyden (changes requested):

zloyden (approved):

PR #111741, opened 2022-03-02
cname trackers #31 #32

slavaleleka:
https://github.com/AdguardTeam/cname-trackers/issues/31

https://github.com/AdguardTeam/cname-trackers/issues/32


SpywareFilter/sections/cname_trackers.txt
```diff
@@ -62,6 +65,9 @@
 ! Lead Forensics : ghochv3eng.trafficmanager.net disguised trackers
 ! not included
 !
+! NameSilo : ab1n.net disguised trackers
+!#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/namesilo.txt
```
AdamWr, line 69: It looks like a list of ad servers, so maybe it would be better to add it to Base filter?

SpywareFilter/sections/cname_trackers.txt
```diff
@@ -44,6 +44,9 @@
 ! Eulerian : eulerian.net disguised trackers
 !#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/eulerian.txt
 !
+! Extole : extole.com disguised trackers
+!#include https://raw.githubusercontent.com/AdguardTeam/cname-trackers/master/trackers/extole.txt
```
AdamWr, line 48: It contains `share.vimeo.com` which seems to be used for referrals - `https://share.vimeo.com/website_footer`, so I guess we should exclude it.
Probably the same for some other `||refer.` rules, so I'm not sure if it should be added.

slavaleleka: I'm not sure as well

@AdguardTeam/filters-maintainers could you please test it for a while?

AdamWr (approved):

Alex-302 (approved):

zloyden (approved):

PR #111783, opened 2022-03-02
Fixed exxen.com for uBlock Origin

bkrucarci:
Problem explained, please read comments as well.
https://github.com/uBlockOrigin/uAssets/issues/11945


AdamWr:
As far as I understand, it works fine with AdGuard, but doesn't with uBlock.
If so, I would prefer to exclude this rule from filters for uBlock than remove it.

bkrucarci:
@AdamWr Can you move the **exxen.com##+js(nofab)** filter to Turkish filter. It is Turkish website, they are right. Turkish filter is default for uBO. if you can move the rule to Turkish filter then they can fix it.
After all, both AdGuard and uBO users will be happy

PR #111911, opened 2022-03-04
Update #90885 - move rules for revolving adservers

Yuki2718:

***Description***:
* **Current behaviour**: 

I don't understand why they're in Annoyances, and it seems my comment in #90885
> /ntfc.php?$script in Base

was wrong, maybe was confusing with uBlock filters. For `.pro/v2/a/push/js/$third-party`, visit [NSFW] `h2porn.com` and you'll see `http://ca.clcknads.pro/v2/a/push/js/37050`. For `https://publicwww.com/websites/%22ntfc.php%22/` and you see a lot e.g. `youtubemp3.to` calls `https://pusherism.com/ntfc.php?p=1733642` blocked by EL.

</details><br/>

* **Expected behaviour**: 

They should be in base to block revolving adservers.

</details><br/>

Visit the site and see requests.

***System configuration***

NA

**Filters:**

NA

### DIFF
--- a/AnnoyancesFilter/sections/push-notifications.txt
+++ b/AnnoyancesFilter/sections/push-notifications.txt
@@ -33,8 +33,6 @@ abendzeitung-muenchen.de###webpush-app
 ||karuna4u.com^$third-party
 ||pushofferpro.com^$third-party
 ! uCoz
-.pro/v2/a/push/js/$third-party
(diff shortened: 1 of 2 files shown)


Alex-302 (approved):

Alex-302:
Yes, it seems push notifications ads/spam.

PR #112003, opened 2022-03-05
rule-maintenance - 7 - old

SKEIDs:
~~`togetter.com###appdl_header`~~, `mdpr.jp##.p-articleComment__buttonArea` are duplicated, is this a problem?

https://github.com/SKEIDs/AdguardFilters/blob/4d4247b8bcd72830b61f9973707d6c76ea3bb559/AnnoyancesFilter/sections/mobile-app.txt#L709


gork7777:
@SKEIDs If there's duplicated rule, it should be removed. Just one rule is enough.

SKEIDs:
> @SKEIDs If there's duplicated rule, it should be removed. Just one rule is enough.

Okay.

PR #112096, opened 2022-03-06
Fix search.naver.com

piquark6046:
https://github.com/AdguardTeam/AdguardFilters/issues/112077
https://github.com/AdguardTeam/AdguardFilters/issues/112089

### How to reproduce
1. Visit [here](https://search.naver.com/search.naver?where=nexearch&sm=top_hty&fbm=0&ie=utf8&query=ad) with AdGuard for Windows.
2. Click the highlighted element.
    <details>
    <summary>Screenshot</summary>

   Screenshot1
 
    </details>

### DIFF
--- a/SpywareFilter/sections/specific.txt
+++ b/SpywareFilter/sections/specific.txt
@@ -52,7 +52,9 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com|bolgehastanesi.com|ewtc.de
 !
+!#if ( !adguard_app_windows )
 ||search.naver.com/p/crd/rd
+!#endif
 ||tpucdn.com/js/ga-*.js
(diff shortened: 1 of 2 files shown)


SpywareFilter/sections/specific.txt
```diff
@@ -52,7 +52,9 @@
 !
 /analytics.js$domain=bataryapil.com|journey.com.tr|freexcafe.com|rusvideos.tv|nakubani.ru|lun.com|songsara.net|partifabrik.com|newsmax.com|bolgehastanesi.com|ewtc.de
 !
+!#if ( !adguard_app_windows )
 ||search.naver.com/p/crd/rd
+!#endif
```
AdamWr, line 57: The same issue occurs on Android (I suppose that the same is with AdGuard for Mac). If I'm not wrong, it's not blocked by extensions, because extensions do not block main document.
So, if it's used only as "document" then I think that it would be better to remove this rule.

AdamWr (approved):

PR #112453, opened 2022-03-10
Fix #112452 www.pcking.de

Morku:
Fix #112452


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -119,6 +119,7 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+pcking.de###es_cookie_solution
```
zloyden, line 122: It looks like we can block it as a request `||pcking.de/_javascripts/cookiesolution/`

Morku (commented):
Update

zloyden (approved):

PR #112574, opened 2022-03-11
Fix twitter.com popup

piquark6046:
### How to reproduce
Just scroll the website down at any thread.

---

https://github.com/AdguardTeam/AdguardFilters/issues/109142


piquark6046:
Example: https://twitter.com/github/status/1425505817827151872?cxt=HHwWgMC4jaP_tMgnAAAA

AnnoyancesFilter/sections/popups.txt
```diff
@@ -2188,7 +2190,6 @@ siberkalem.com#$#body { overflow: visible !important; }
 tweaktown.com##.newsletter-overlay
 cyberleninka.ru##body > div.info
 stanleyblackanddecker.com###popup-salesforce-signup
-twitter.com#?##layers > div[class] > div[class] > div[class] > div[class]:has(> div[class] > div[class] > div[class] > div[class] > div[class] > a[href="/login"])
```
Alex-302, line 2191: This rule is still required
image

Alex-302 (changes requested):

AnnoyancesFilter/sections/popups.txt
Alex-302: Big popup cannot reproduce.
@AdguardTeam/filters-maintainers Please take a look. Maybe can be blocked by scriptlet(`prevent-addEventListener` on scroll event, or something else)

piquark6046: https://github.com/AdguardTeam/AdguardFilters/pull/112574#discussion_r824813341
Okay. I pushed.

AnnoyancesFilter/sections/popups.txt
```diff
@@ -119,6 +119,8 @@ sodapdf.com##.be-tooltip
 ! Specific sites
 !######### Subscription popups add to subscriptions.txt #########
 !
+twitter.com#?##layers > div[class] > div[class] > div[class] > div[class] > div[class] > div[class] > div[class][tabindex][role]:has(> div[class][data-testid] > div[class] > div[class] > div[class] > div[class] > div[class] > div[class] > div[class] > span > span[class]:contains(Log in))
```
Alex-302, line 122: The first rule does not applied on my end.
What about this?
```
twitter.com#$?#div[data-testid="sheetDialog"]:upward(div[role="group"]) { display: none !important; }
twitter.com#$?#html { overflow: auto !important; padding-right: 0 !important; }
```
If you have account, please test with/without account in new private tabs.

Alex-302 (commented):
Better move below to the rules with comments, and add a comment about purpose of these rules

Alex-302 (approved):

PR #112830, opened 2022-03-14
do not include logly.co.jp disguised trackers

slavaleleka:
https://github.com/AdguardTeam/cname-trackers/issues/35


Yuki2718 (commented):
If only these two, we can probably enable the list safely.

Alex-302 (approved):

zloyden (approved):

AdamWr (approved):

PR #112902, opened 2022-03-14
arzum.com.tr

gork7777:
issue: #112898


SpywareFilter/sections/whitelist.txt
```diff
@@ -26,6 +26,8 @@
 !#endif
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/112898
+@@||googletagmanager.com/gtm.js$domain=arzum.com.tr
```
AdamWr, line 30: It seems that in this case a `$redirect` rule works fine, so I think that it's preferable.
```
||googletagmanager.com/gtm.js$script,redirect=googletagmanager-gtm,domain=arzum.com.tr
```
I will add it.

AdamWr (approved):

PR #113358, opened 2022-03-20
rule-maintenance - 9

SKEIDs:
AnnoyancesFilter/sections/subscriptions.txt


AnnoyancesFilter/sections/subscriptions.txt
```diff
@@ -583,7 +574,6 @@ netology.ru##a[href^="https://tgclick.ru/netology_ru/"]
 medizinfuchs.de##.NewsletterInlineContent
 reidl.de##.start-get-newsletter
 cashfloat.co.uk##.social-col
-internet-technologies.ru##.subscription-social
```
zloyden, line 586: Why the rule was removed? It's still needed, actually.

zloyden (changes requested):

AnnoyancesFilter/sections/subscriptions.txt
zloyden: I see, ok )

zloyden (approved):

PR #113722, opened 2022-03-24
oceans-nadia.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/112440


Yuki2718 (changes requested):
Maybe these and the new rules should be under NOT_OPTIMIZED
```
oceans-nadia.com###js-getAppHeader
oceans-nadia.com##div[id^="app-banner"]
oceans-nadia.com##.l-footer__app
```

Yuki2718:
@ameshkov I wonder if it's possible to exclude the whole mobile-app.txt from optimization, given most of enires in the list are for mobile env where optimized list is default.

Alex-302 (approved):

PR #113796, opened 2022-03-25
Fix for copays.org (charity)

Gitoffthelawn:
Fixes https://copays.org/stay-connected/
(this is a charity that helps people get life saving meds)

### DIFF
--- a/AnnoyancesFilter/sections/subscriptions.txt
+++ b/AnnoyancesFilter/sections/subscriptions.txt
@@ -33,7 +33,7 @@ turkuazgazetesi.net,yenialanya.com,mansetaydin.com,seskocaeli.com,kocaelibarisga
 ||embeds.fanmatics.com^$third-party
 ||cp.unisender.com/*/form-widget-loader.js$third-party
 ##.widget_blog_subscription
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/subscriptions.txt
```diff
@@ -33,7 +33,7 @@ turkuazgazetesi.net,yenialanya.com,mansetaydin.com,seskocaeli.com,kocaelibarisga
 ||embeds.fanmatics.com^$third-party
 ||cp.unisender.com/*/form-widget-loader.js$third-party
 ##.widget_blog_subscription
-~nothing.tech,~mega-debrid.eu,~curvegames.com,~news.un.org,~norwichzen.org.uk,~dereferer.me###mc_embed_signup
+~copays.org,~nothing.tech,~mega-debrid.eu,~curvegames.com,~news.un.org,~norwichzen.org.uk,~dereferer.me###mc_embed_signup
```
zloyden, line 36: What's the reason to unblock it? It's still a subscription. User can disable AdGuard Annoyances.

zloyden: I see that this subscription is only here https://copays.org/stay-connected/, so it's ok to unblock it.

zloyden (approved):

PR #114813, opened 2022-04-06
rule-maintenance - 21

SKEIDs:
AnnoyancesFilter/sections/popups.txt

### DIFF
--- a/AnnoyancesFilter/sections/popups.txt
+++ b/AnnoyancesFilter/sections/popups.txt
@@ -3047,7 +3047,6 @@ discoverskills.com##.et_bloom_popup
 lamoda.ru##.unipop__container
 kicksusa.zendesk.com,ibe.kr,sneakerhead.com,support.newbalance.com##.zEWidget-launcher
 shoes.com###_GUARANTEE_SealSpan
-bigfishgames.de,adapostasi.com##div[id^="simplemodal-"]
 monday.com##.exit-popup-overlay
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/sections/popups.txt
```diff
@@ -429,7 +429,6 @@ blokt.com##div[id^="coinzilla_expandable_widget"]
 cp.glico.jp###sec_modal_banner
 gd.ru#$#.paywall[data-window-name-for-ga] { display: none !important; }
 gd.ru#$#.megaShadow { display: none !important; }
-gd.ru#$#body[style^="overflow:"] { overflow: auto !important; }
```
Alex-302, line 432: better do not ungroup CSS rules. In its current variant, it is more convenient.

Alex-302: Please revert CSS changes an I will merge.

zloyden: @SKEIDs you can sort simple rules, but `#$#` is not necessary here, because we have some rules for each site that are easy to understand or change later on.

zloyden (changes requested):

AnnoyancesFilter/sections/popups.txt
SKEIDs: thank you for advice!

PR #114815, opened 2022-04-06
rule-maintenance - 22

SKEIDs:
AnnoyancesFilter/sections/popups.txt


Alex-302:
In previous commend explained why it is not best change.
But if all rules from group are the same for few sites, they can be corted.

PR #115459, opened 2022-04-13
www.news.com.au

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/115340


SpywareFilter/sections/whitelist.txt
```diff
@@ -37,6 +37,8 @@
 @@||googletagmanager.com/gtm.js$domain=security-demo.extrahop.com
 ||googletagmanager.com/gtm.js$script,xmlhttprequest,redirect=noopjs,domain=security-demo.extrahop.com,important
 extrahop.com#%#//scriptlet('prevent-setTimeout', 'ad blocker')
+! https://github.com/AdguardTeam/AdguardFilters/issues/115340
+@@||tags.news.com.au/prod/$domain=news.com.au
```
Alex-302, line 41: On my end `@@||tags.news.com.au/*/MediaSDK.min.js$domain=news.com.au` is enough, but player is broken by Easyprivacy.

zloyden: I have the same, so we don't need the exception rule.

PR #115479, opened 2022-04-13
gooday.nikkei.co.jp

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/115432

I think that we shouldn't block banners from screenshot in the issue, because they leads on internal pages

Снимок экрана 2022-04-13 в 19 32 00

</details><br/>


Alex-302 (approved):

Alex-302:
Ok.

PR #115819, opened 2022-04-17
www.bataryapil.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/115662

removed domains from `/analytics.js` , that blocked by `cdn.ticimax.com/*/analytics.js`


Alex-302:
In general this is not a mistake, but well, ok.

Alex-302 (approved):

PR #115862, opened 2022-04-17
ing.com

gork7777:
issue: #115750

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -120,6 +120,8 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+ing.com#$#body { overflow: auto !important; }
+ing.com#$#.global-overlays { display: none !important; }
 ||googletagmanager.com/gtm.js$domain=advcash.com
 fuvest.br###idInformacaoCookie
 imobie.jp##.fixedbot


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -120,6 +120,8 @@ techadvisor.co.uk,forum.notebookreview.com,computerworld.com,skysports.com#$#bod
 !
 !
 !
+ing.com#$#body { overflow: auto !important; }
+ing.com#$#.global-overlays { display: none !important; }
```
Alex-302, line 124: It is used on dev domain only. The rule looks wide and dangerous.

Alex-302 (changes requested):

AnnoyancesFilter/sections/cookies_specific.txt
gork7777: Can we change that like this?
```
developer.ing.com#$#body { overflow: auto !important; }
developer.ing.com##.global-overlays__backdrop
developer.ing.com##.global-overlays__overlay
```

Alex-302: just change domain to developer.ing.com in original rules.

> Can we change that like this?

No. Please do not mix element hiding rules with CSS, which applied to one problem(for example, popup + scroll unblocking).

Alex-302: Added directly.

PR #116409, opened 2022-04-24
Remove duplicates or rules already added to easylist

SpicySpaceman:

***Description***: Removing rules which are duplicates in the AdGuard filter itself or already added by easylist. Let me know if there's any reason to keep duplicates so I will stop searching for them ;)


slavaleleka:
I suppose [FilterCompiler](https://github.com/AdguardTeam/FiltersCompiler) should remove duplicated while filters build

zloyden (approved):

PR #116412, opened 2022-04-24
www.fullhdizle.com

gork7777:
issue: #116411


Alex-302 (approved):
also added `fullhdizle.com##.kapat-div`

PR #116598, opened 2022-04-26
www.instiz.net - Annoyances 

green1052:
URL: https://www.instiz.net/pt/7138479

IMAGE: 
image


piquark6046 (changes requested):
It seems that an extended CSS injection rule does not required.

green1052 (commented):
fix

Alex-302:
I see no difference. What the purpose of the rule?
(btw. `www` usually not required)

green1052:
its remove water mark

before:
image

after:
image

Alex-302 (approved):

PR #116854, opened 2022-04-28
Update Turkish adservers_firstparty.txt

bkrucarci:
"^$third-party" rules moved to adservers.txt
The rules alphabetically ordered.
Please see the other PR: https://github.com/AdguardTeam/AdguardFilters/pull/116855


Alex-302 (approved):

Alex-302:
In general, there is no need to do this. At least not not by hand. We need new sorting script for that.

bkrucarci:
> In general, there is no need to do this. At least not by hand. We need new sorting script for that.

I did it with Notepad++ just one-click 👍🏻 
Thanks for merging the PR's.

PR #117165, opened 2022-05-01
www.yaani.com.tr

gork7777:
issue: #117152


AdamWr:
I think that in case like this, it's better to modify/remove problematic rule than add exception.
I have fixed it directly.

PR #117328, opened 2022-05-03
transition-news.org

spiri-leo:
Issue: #117268


Alex-302:
Rule is ok, but the site has too few visitors.

PR #117492, opened 2022-05-04
obsolete filter

JobcenterTycoon:
All domains are completely dead


Alex-302:
Did you take into account that for some advertising systems only subdomains are resolved?(f.e. `ads.example111.com` used for ads, but `example111.com` has no DNS records and looks like dead)

Alex-302:
I don't trust to services like that:)

JobcenterTycoon:
Ok and how you check a domain to be sure its really dead ?

Alex-302:
I do not check)

zloyden (approved):

PR #117663, opened 2022-05-05
obsolete filter part 2

JobcenterTycoon:
Same as: https://github.com/AdguardTeam/AdguardFilters/pull/117492


Alex-302:
It is really not the best idea to remove URL blocking rules with dead domains.
Example:
tracking domain `vgs-collect-keeper.apps.verygood.systems` 
it can be blocked by `||verygood.systems^` or `||apps.verygood.systems^`
In this case DNS checker will report about dead domain - not resolved by DNS. But `vgs-collect-keeper.apps.verygood.systems` - working, and tracking will be unblocked after removing "obsolete" rule.

JobcenterTycoon:
It does not, it checks if the domain is really avaiable (to rent)

Alex-302:
What site do you use for checking?

JobcenterTycoon:
`https://www.eurodns.com/de/bulk-search` exacly the same as https://github.com/AdguardTeam/AdguardFilters/pull/117492

Alex-302:
I don't trust to this service. Need time to check.

zloyden (approved):

Alex-302 (approved):

PR #117819, opened 2022-05-06
General rules for telephone ads

terlancenablari:
I think it is already time to create general rules for extensions such as info, com and net which are used in porn sites. Otherwise, we should add a new spesific rule every time and it is time wasting stuff.


Alex-302:
I will do it later with a bit different rules.

PR #118029, opened 2022-05-08
İn alphabetical order (css_extended.txt)

terlancenablari:


Alex-302:
Thanks, but this is literally useless work:) Sorting should take a different approach.

PR #118721, opened 2022-05-14
primevideo.com Referrer

bkrucarci:
After a login from amazon.com, please merge it.


Alex-302:
Not sure about it. Literally track nothing. But some sites are not working correctly, when similar parameter is removed.

PR #119413, opened 2022-05-21
Dead domains (1)

mrrtarlan:
bifilmizlesene.net
limitsizfilmizle.net
rexfilmizle.com
favorifilm.com
hdfilmizle3.com
filmihdizlet.com
fullizlefilmizle.com
hdcehennem.com
cilekfilm.com
wfilmizle.net
filmkusu.org (redirects to filmizlemetv.com which has active rules in filter; so rules related to this domain are useless)
kalitelifilmizle.net (site content are different, maybe site view changed)
ifilmizle.org
dafflix.com
hdfilmcehenemi.net
filmoyunindir.com
erotikfilms.net (this domain redirects to erotikfilmizle1.com, so rules related to this domain are useless)
erotiksexfilmizle.com
filmsokagi.org
full1080pfilmizle.com

### DIFF
--- a/TurkishFilter/sections/specific.txt
+++ b/TurkishFilter/sections/specific.txt
@@ -8,9 +8,9 @@
 !
 ! Images hosting. Used for advertising on many sites
 !+ NOT_OPTIMIZED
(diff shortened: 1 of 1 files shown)


gork7777:
Some domains may be located in files other than the language specific file. That's why you shouldn't just look at the language specific file.

mrrtarlan:
> Some domains may be located in files other than the language specific file. That's why you shouldn't just look at the language specific file.

is there any way to edit rules from one place? as you said, currently i can only edit just one section and that's why there should be rules in other filters

gork7777:
@mrrtarlan It's better do not make PR's about dead domains, you can just report issues about dead domains and devs will check and remove them.

krystian3w (approved):

zloyden (approved):
For such cases, better do it with different commits next time.
For instance
```
example.com - dead
example1.com - dead
```

PR #120540, opened 2022-06-01
fix anti adblock on francaisfacile.net

JobcenterTycoon:
### Please include a summary of the change and which issue is fixed.

Adblock get detected `https://francaisfacile.net/`

---

##### To avoid invalid pull requests, please check and confirm following checkboxes:

##### If the problem does not fall under any category that is listed here, please write a comment below in corresponding section

  - [x] AdGuard gets detected on a website;
  - [ ] Filters maintenance.

##### Enter the issue address:

no issue in the issue tracker

##### If possible, a screenshot of a page or application should not be cropped too much. Otherwise, it is not always clear where the element is located.

<details>
Unbenannt
</details>

### Terms

  - [x] By submitting this issue, I agree that pull request does not contain private info and all conditions are met.

### DIFF
--- a/FrenchFilter/sections/antiadblock.txt
+++ b/FrenchFilter/sections/antiadblock.txt
@@ -17,6 +17,8 @@ capital.fr#@#.plainAd
 !
 !
 !
+! https://www.reddit.com/r/uBlockOrigin/comments/v2b6l4/asking_to_unblock_the_adblocker/
+francaisfacile.net#%#//scriptlet("prevent-addEventListener", "DOMContentLoaded", "Adblock")
 ! https://github.com/AdguardTeam/AdguardFilters/issues/118024
 @@||parlons-basket.com/wp-content/*/js/prebid-ads.js
 parlons-basket.com#%#//scriptlet("set-constant", "canRunAds", "true")


FrenchFilter/sections/antiadblock.txt
```diff
@@ -17,6 +17,8 @@ capital.fr#@#.plainAd
 !
 !
 !
+! https://www.reddit.com/r/uBlockOrigin/comments/v2b6l4/asking_to_unblock_the_adblocker/
+francaisfacile.net#%#//scriptlet("prevent-addEventListener", "DOMContentLoaded", "Adblock")
```
zloyden, line 21: We already have the rules for this. FIxed https://github.com/AdguardTeam/AdguardFilters/commit/c6c13d41d6f145518b9c5fea969a5ffb6639362d

JobcenterTycoon:
@zloyden i see there are many sites with this script. The script only checks 4 cosmetic filter so what about to change the 4 filter to

```
##.adbadge:not(#filter_ads_by_classname)
##.ad-placeholder:not(#filter_ads_by_classname)
##.BannerAd:not(#filter_ads_by_classname)
##.ads_1:not(#filter_ads_by_classname)
```
to fix the anti adblock globally?

zloyden:
As I can see, those rules from `Easylist`, so we can't change them directly. For now current solution is enough.
@AdamWr what do you think?

AdamWr:
If I'm not wrong, this script is also checking if `pagead2.googlesyndication.com/pagead/js/adsbygoogle.js` is blocked, so we would need also to add generic redirect rule. And if there are exceptions for these element hiding rules, then we will have to modify them too.

What if we would just add:
```
##div[id^="chp_ads_block"][class*="chp_ads_blocker_detector"]
```
as a generic rule.
It should be enough in most cases (sometimes scroll is blocked, but it seems that it's not so often).

zloyden:
First, it should be tested, but I agree.
We can disable current rule and add the rule above to `! chp_ads_blocker` section.

JobcenterTycoon:
Works fine for the set of domains i got from publicwww

```
https://bluebottle.idv.tw/
https://www.elreporte.com.uy
https://www.flashfile.in/
https://www.sempredirebanzai.it/
http://vectogravic.com/
https://rinconpsicologia.com/
https://francaisfacile.net/
https://hdking.vip/
```

zloyden:
> Works fine for the set of domains i got from publicwww

Actually, not everywhere from the list above -> [francaisfacile.net](https://francaisfacile.net), but added it.

PR #121441, opened 2022-06-10
Fix some manga websites ads

SKEIDs:
#121373, #121372, #121371, #121370, #121369, #121368, #121367, #121366 (ss1), #121365 (ss1)


Yuki2718 (commented):
@Alex-302 @zloyden Probably better to make it generic.

PR #121782, opened 2022-06-14
update pull_request_template

kaprielov:
Added description of the filter-pr-checker steps

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -32,6 +32,14 @@
 
 Example: <https://github.com/AdguardTeam/AdguardFilters/issues/100000>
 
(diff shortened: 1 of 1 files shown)


Alex-302:
The most of pull requests are created for existing issues. We ask to link them.
If no such issue - user can send report from AdGuard directly and attach the link.

maximtop (changes requested):
Let's wait until @ameshkov approves the action. Also, we should connect action in this PR as well.

ameshkov (approved):

ameshkov (changes requested):
Well, according to the last task, we shouldn't force any special rules for the PR text

PR #121860, opened 2022-06-15
automexico.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/121548

### DIFF
--- a/SpywareFilter/sections/general_url.txt
+++ b/SpywareFilter/sections/general_url.txt
@@ -46,7 +46,8 @@
 ! Facebook
 /js/fb_tracking.js
 /fbpix-events-
-/fbevents.
+/fbevents.js
+/fbevents.min.js
 !
 ! Cloudflare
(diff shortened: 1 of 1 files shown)


SpywareFilter/sections/general_url.txt
```diff
@@ -46,6 +46,7 @@
 ! Facebook
 /js/fb_tracking.js
 /fbpix-events-
+/fbevents.
```
zloyden, line 49: Actually, there is a script https://static.automexico.com/scripts/fbevents.min.js.gz?v=ea5a2775-d0f4-4580-9240-cda0ecaf6909.
So, you also should add 
```
/fbevents.min.js
```

zloyden (changes requested):

SpywareFilter/sections/general_url.txt
zloyden: Another option is to create two rules
```
/fbevents.js
/fbevents.min.js
```

zloyden (approved):

PR #123734, opened 2022-07-07
www.tomshw.it

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/123443

Снимок экрана 2022-07-07 в 13 13 28

</details><br/>

### DIFF
--- a/AnnoyancesFilter/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/sections/cookies_specific.txt
@@ -138,6 +138,8 @@ voici.fr,the-sun.com,thesun.ie,capital.fr,ladbible.com,hitc.com,dualshockers.com
 !
 livoo.fr#$##tarteaucitronAlertBig { display: none!important; }
 livoo.fr#$#html { overflow: auto!important; }
+tomshw.it#$#html, body { overflow: auto !important; }
+tomshw.it#$##iubenda-cs-banner { display: none !important; }
 handelsbanken.co.uk##.shb-cookie-consent__wrapper
 danskebank.co.uk#$#body { overflow: auto !important; }
 danskebank.co.uk#$#.cookie-consent-banner-modal { display: none !important; }


AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -138,6 +138,8 @@ voici.fr,the-sun.com,thesun.ie,capital.fr,ladbible.com,hitc.com,dualshockers.com
 !
 livoo.fr#$##tarteaucitronAlertBig { display: none!important; }
 livoo.fr#$#html { overflow: auto!important; }
+tomshw.it#$#html, body { overflow: auto !important; }
```
zloyden, line 141: 1. We do not add more than one selector to a single rule.
2. We already have the rules that change `overflow`.

AnnoyancesFilter/sections/cookies_specific.txt
```diff
@@ -138,6 +138,8 @@ voici.fr,the-sun.com,thesun.ie,capital.fr,ladbible.com,hitc.com,dualshockers.com
...
+tomshw.it#$##iubenda-cs-banner { display: none !important; }
```
zloyden, line 142: Also add an exception for content blockers.

zloyden (changes requested):

zloyden:
Actually, it opens fine on my end. Maybe the current JS rule does not apply on Firefox? In that case, we should not fix anything, but we can give the rule to user.

AdamWr:
Comments do not open if cookie notice is not accepted, like user described it here - https://github.com/AdguardTeam/AdguardFilters/issues/121221
I still can reproduce this issue.
With apps and extensions for Chrome/Edge this cookie is automatically closed, but extensions for Opera/Firefox cannot download JS rules, so cookie notice is not hidden.

zloyden:
> extensions for Opera/Firefox cannot download JS rules

I've answered [here](https://github.com/AdguardTeam/AdguardFilters/issues/123443#issuecomment-1177377476).

PR #123756, opened 2022-07-07
www.canalplus.com

### LINKED ISSUE #116568: www.canalplus.com
### Issue URL (Ads)
 [https://www.canalplus.com/divertissement/maries-au-premier-regard/h/7643336_50035](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.canalplus.com%2Fdivertissement%2Fmaries-au-premier-regard%2Fh%2F7643336_50035)

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
AdGuard product: | AdGuard Browser extension v4.0.163
Browser: | EdgeChromium
Stealth mode: | disabled
Filters: | <b>Ad Blocking:</b><br/>AdGuard Base<br/><br/><b>Other:</b><br/>Filter unblocking search ads and self-promotion<br/><br/><b>Language-specific:</b><br/>AdGuard French, <br/>AdGuard German

### LINKED ISSUE #123526: www.canalplus.com
### Issue URL (Ads)
 [https://www.canalplus.com/divertissement/l-academie-des-gateaux-de-cyril-lignac/h/18911550_50035](https://adguardteam.github.io/AnonymousRedirect/redirect.html?url=https%3A%2F%2Fwww.canalplus.com%2Fdivertissement%2Fl-academie-des-gateaux-de-cyril-lignac%2Fh%2F18911550_50035)

### Screenshots
<details>

  Screenshot 1
</details>

### System configuration

Information | value
--- | ---
AdGuard product: | AdGuard Browser extension v4.0.181
Browser: | EdgeChromium
Stealth mode: | disabled
Filters: | <b>Ad Blocking:</b><br/>AdGuard Base<br/><br/><b>Privacy:</b><br/>AdGuard Tracking Protection<br/><br/><b>Social Widgets:</b><br/>AdGuard Social Media<br/><br/><b>Annoyances:</b><br/>AdGuard Annoyances<br/><br/><b>Language-specific:</b><br/>AdGuard French, <br/>AdGuard German, <br/>EasyList Italy

contribucious:
# Creating the pull request

This PR ensures that video ads never appear on `canalplus.com`.

- [X] This is not an ad/bug report;
- [X] My code follows the [guidelines](https://github.com/AdguardTeam/AdguardFilters/blob/master/CONTRIBUTING.md) and [syntax](https://kb.adguard.com/general/how-to-create-your-own-ad-filters) of this project;
- [X] I have performed a self-review of my own changes;
- [X] My changes do not break web sites, apps and files structure.

- [X] Missed ads or ad leftovers; &nbsp;_(Possibly, in rare instances.)_
- [ ] AdGuard gets detected on a website; &nbsp;_(Currently undetected but improves/reinforces that side as well.)_
- [X] Filters maintenance. &nbsp;_(Unplanned reorganization included, which should have been done in a separate commit for easier review, sorry about that.)_

### Enter the issue address

Fixes #123526, fixes #116568

**_1. Your comment_**

Hi. First of all, sorry for the delay.

In short, these two issues can _probably_ be explained either by a glitch (see in particular the fact that two instances of AdGuard are used simultaneously on one of the two issues, both of which seem to me to come from the same person by the way) or by the fact that some premium content (coming from their partnership with _6play_ here) would _potentially_ not use `.featPlayerAds`. Although that doesn't seem to be the case.

Anyway, the current solution, although never faulty in my tests, is a bit fragile on a glitch level because it must be correctly defined once the DOM of the page is loaded _(document)_ but before its use ~ 2 sec. later in `main.[…].js` _(script)_.

Therefore, for the possible rare case where my `set-constant` line wouldn't react fast enough, added:
- its `$document,replace=[…]` version (for AG products supporting it);
- a deletion of the following JSON content via the `json-prune` scriptlet, where `*` is any ad provider name:
  - `ads.*.default` _[binary]_ (to allow to never see [that](https://user-images.githubusercontent.com/4764956/141113868-85bdb3d1-1f24-4ac2-ac20-0000afc560b9.png) in this situation too — English translation: "Advertising: Your content is loading");
  - `ads.*.url` _[string]_ (to prevent contacting any ad provider, currently two present, now or later; with `ads.enableAntiAdBlocking` removed alongside);
  - `ads.enableMidrolls` _[binary]_ (to prevent any possible/future 'OSD' visual stuff showing before trying to show them).

The final `, 'ads'` _(another parameter)_ being for optimization purposes: if I understand it correctly, this will allow in this case for each "JSON data" where the `[ROOT].ads` property is absent, to only look for its presence once instead of 3 times.

**_2. Screenshots_**

<details>

<summary>:camera: Screenshot 1 <strong><em>(click to reveal)</em></strong></summary>

firefox_GMvvbaaA6k

**Source:** `https://player.canalplus.com/one/configs/v2/10/mycanal/prod.json` _(**OnePlayer** is the name of their player)_
… but will also works on all variants visible here (`https://player.canalplus.com/one/configs/v2/10/mapping/mapping.json`), e.g.:
- `https://player.canalplus.com/one/configs/v2/10/mycanalch/prod.json`
- `https://player.canalplus.com/one/configs/v2/10/orange/prod.json`
- `https://player.canalplus.com/one/configs/v2/10/piwi/prod.json` (for kids, where we can see that `ads.enableMidrolls` is already `false` there for example)
- etc.

:bulb: And if you are wondering: `params.enableAd`, either removed or set to `false` via a `$replace` rule, has no influence on _video ads_.

</details>

<details>

<summary>:camera: Screenshot 2 <strong><em>(click to reveal)</em></strong></summary>

firefox_ne8QfUSCNC

_:information_source: If `json-prune` only used, production of this error but does not pose any problem (same result with an adequate **`redirect`** rule [noopvmap-1.0](https://github.com/AdguardTeam/Scriptlets/blob/master/wiki/about-redirects.md#noopvmap-1.0) on `proxy.ads.canalplus-bo.net/[…]` for example, by the way)._

</details>

### Terms

- [X] By submitting this issue, I agree that pull request does not contain private info and all conditions are met

### DIFF
--- a/FrenchFilter/sections/antiadblock.txt
+++ b/FrenchFilter/sections/antiadblock.txt
@@ -57,11 +57,6 @@ chat.babel.com#%#//scriptlet('prevent-setTimeout', 'ShowAdvertising')
 @@||googletagmanager.com/gtag/js$domain=chat.babel.com
 @@||adv.123multimedia.com/AdConfig.aspx$domain=chat.babel.com
 !#endif
-! https://github.com/AdguardTeam/AdguardFilters/issues/102748
-qub.ca#$#.adsbox.ad-banner:not([style="height: 5px; width: 5px; position: absolute; top: 0;"]):not(.blocker-tester + .ad-banner) { display: block !important; }
-qub.ca#$#.pub_300x250.pub_300x250m.pub_728x90.text-ad.textAd.text_ad.text_ads.text-ads.text-ad-links { display: block !important; }
-!+ PLATFORM(ios, ext_android_cb)
-@@||qub.ca^$generichide
 ! https://github.com/AdguardTeam/AdguardFilters/issues/101696
 @@||abcbourse.com/scripts2/prebid.js
 abcbourse.com#$##mywraptest { height: 1px !important; }
@@ -161,10 +156,6 @@ jardiner-malin.fr#%#//scriptlet("set-constant", "wIsAdBlocked", "false")
 @@||cdn-a.yieldlove.com/yieldlove-bidder.js$domain=jardiner-malin.fr
 ! https://github.com/AdguardTeam/AdguardFilters/issues/54079
 bleachmx.fr#%#//scriptlet("abort-on-property-read", "mdpDeBlocker")
-! https://github.com/AdguardTeam/AdguardFilters/issues/53934
-@@||proxy.ads.canalplus-bo.net/web/*/advert/?adInfo=$domain=canalplus.com
-@@||v.fwmrm.net/ad/g/1?*&resp=vmap$domain=canalplus.com
-@@||proxy.ads.canalplus-bo.net/web/*/adserver/info/?adInfo=$domain=canalplus.com
 ! https://github.com/AdguardTeam/AdguardFilters/issues/53875
 @@||leakgaming.fr/js/*/Adsplex-shu.js
 leakgaming.fr#%#//scriptlet("abort-current-inline-script", "$", "!document.getElementById(btoa")
@@ -275,9 +266,6 @@ sante.journaldesfemmes.fr,linternaute.com#$#.AdBox.Ligatus.pub_300x250.pub_300x2
 @@||kweb.r66net.com/getlink|$domain=linternaute.com
 @@||pagead2.googlesyndication.com/pagead/js/adsbygoogle.js$domain=linternaute.com
 !#endif
-! https://forum.adguard.com/index.php?threads/15884/
-@@||static.canal-plus.net/resources/adframe.html?adName=$domain=canalplus.com
-@@||proxy.ads.canalplus-bo.net/web/*/adserver/info/?asset=$domain=canalplus.com
 ! https://forum.adguard.com/index.php?threads/m-rtl-be-adguard-mobile.31398/
 @@||service.videoplaza.tv/proxy/bwtest.jpg$domain=rtl.be
 @@||service.videoplaza.tv/proxy/pulse-sdk-html5^$domain=rtl.be
@@ -318,8 +306,6 @@ illicoporno.com#%#window.is_adblocked = false;
 ! https://forum.adguard.com/index.php?threads/www-culture-generale-fr.30064/
 culture-generale.fr#@#.an-advert-banner
 culture-generale.fr#@#.an-sponsored
-! https://forum.adguard.com/index.php?threads/www-rtlplay-be-tvi-humour-rtl-tvi.29736
-@@||be-rtl.videoplaza.tv/proxy/distributor/$domain=rtlplay.be
 ! https://github.com/AdguardTeam/AdguardFilters/issues/25494
 ledauphine.com###adBlockDetect
 ! https://github.com/AdguardTeam/AdguardFilters/issues/22988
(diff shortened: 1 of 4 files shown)


Alex-302 (approved):

contribucious:
:v: :relaxed:

contribucious:
## Small correction (@Alex-302)
&nbsp;
> same result with an adequate `redirect-rule` [noopvmap-1.0](https://github.com/AdguardTeam/Scriptlets/blob/master/wiki/about-redirects.md#noopvmap-1.0) on `proxy.ads.canalplus-bo.net/[…]`

:arrow_right_hook: Naturally, it should have read "with an adequate `redirect` rule" instead of "with an adequate `redirect-rule`" here.

:heavy_check_mark: Simple drafting error in my post but however, to be sure, I rechecked the rule (kept) that I had used during my tests [ereyesterday](https://www.quora.com/What-do-you-call-the-day-before-yesterday)([*](https://en.wiktionary.org/wiki/ereyesterday)) and it was correct:
`||proxy.ads.canalplus-bo.net/web/*/advert/?adInfo=$xmlhttprequest,other,redirect=noopvmap-1.0,domain=canalplus.com`

contribucious:
## Side related dev question (@slavaleleka)
&nbsp;
Hello Slava. Small question for you.

**Context is this rule (and [this scriptlet](https://github.com/AdguardTeam/Scriptlets/blob/fdef488c1f529ce85f87d441eafbf0422b3d8336/src/scriptlets/json-prune.js)):**
`canalplus.com#%#//scriptlet('json-prune', 'ads.*.default ads.*.url ads.enableMidrolls', 'ads')`

**Remark (from above) is:**
> The final `, 'ads'` _(another parameter)_ being for optimization purposes: if I understand it correctly, this will allow in this case for each "JSON data" where the `[ROOT].ads` property is absent, to only look for its presence once instead of 3 times.

:arrow_right_hook: Do you confirm this? If so, wouldn't it be ideal that this scriptlet detects at the very beginning of its processing if all the values of the main parameter have the same base and if so, would set by itself (internally) this second parameter to 'ads' (to take this example), allowing me in this case to omit it while remaining optimized?

**Also:** if I needed a second condition **abc**, I would have written 'ads abc' currently. With this system idea implemented, I would only need to write 'abc' and internally, self-optimized, this would be used: 'ads abc'.

:bulb: **What do you think about this?**

Alex-302:
@contribucious Ok, added
`||proxy.ads.canalplus-bo.net/web/*/advert/?adInfo=$xmlhttprequest,other,redirect=noopvmap-1.0,domain=canalplus.com,important`

contribucious:
@Alex-302 Sorry, I misspoke. It was not clear.

This was following a correction in the "Screenshot 2" section (since corrected). And that was just a rule used in my local tests (_a useless rule_ thanks to the `json-prune` system [… `ads.*.url` …] which prevents this request from being made — scriptlet which is supported by more products than `$redirect`).

:arrow_right_hook: In fact, rule reported just to say that, versus the **used** `json-prune` system, this alternative method also produced the same _non-problematic_ error visible in _Firefox Developer Tools_ on this second screenshot so that it was OK/normal and not to worry about it.

:heavy_check_mark: But **we can leave it however now**, it will serve as an interesting fallback to the `json-prune` system (if they ever change the name of their JSON properties one day). :thumbsup: :wink:

PR #125124, opened 2022-07-22
add mobile app filters for mobile.mi.co.id

proletarius101:

***Description***:
* **Current behaviour**: 

image
</details><br/>

* **Expected behaviour**: 

image

</details><br/>

- Visit https://mobile.mi.co.id/id/ on mobile

***System configuration***

**Filters:**

AdGuard Annoyances filter


AnnoyancesFilter/MobileApp/sections/mobile-app_specific.txt
```diff
@@ -1982,6 +1982,7 @@ ozon.ru#$#body { overflow: visible!important; }
 ozon.ru#$#html { overflow: visible!important; }
 newzit.com#$#div[class*="hasSmartBanner_"] > div[class^="banner_"] { display: none !important; }
 newzit.com#$#div[class*="hasSmartBanner_"] { padding-top: 0 !important; }
+mobile.mi.co.id##.header-storeapp
```
Sergey-Lyapin, line 1985: Please add new rules on top of the list.

PR #126425, opened 2022-08-06
soft98.ir

MasterKia:
Address: https://github.com/AdguardTeam/AdguardFilters/issues/126252

Related: https://github.com/uBlockOrigin/uAssets/pull/14278

I'll send pull request everyday to update the filters.

cc @zloyden @Yuki2718

PR #126697, opened 2022-08-09
Update soft98

MasterKia:
https://github.com/AdguardTeam/AdguardFilters/pull/126425

### DIFF
--- a/EnglishFilter/sections/foreign.txt
+++ b/EnglishFilter/sections/foreign.txt
@@ -6452,19 +6452,29 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
 zoomit.ir#%#document.cookie = "BnShowNum=4; path=/;";
 ! soft98.ir - Persian Blocker rules (daily fix)
 !#if (!adguard_app_windows && !adguard_app_mac && !adguard_app_android && !adguard_ext_android_cb)
-soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="insta"]) { remove: true; }
+soft98.ir#%#//scriptlet("remove-class", c6e1y2, *, stay)
 soft98.ir#$?#[href*="linkdoni.soft98.ir"] { remove: true; }
-soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/aside/section[*]/header) { remove: true; }
-soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/aside/div/section/header) { remove: true; }
+soft98.ir##[src*="file.soft98.ir"]{color: #437593 !important; background-color: #437593 !important; border: #437593 !important;}
+soft98.ir##[src*="cdn.hostdl.com"]{color: white !important; background-color: white !important; border: white !important;}
+soft98.ir,~forum.soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="98"]){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/section[*]/header){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div/div/section[*]/header){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/section[*]/div):has(img[src*="cdn.hostdl.com"]){ color: white !important; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/div/section/header){ remove: true; }
 soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/div/section/div){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div/div/div/section){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/main/article[*]/span){ remove: true; }
+soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/main/article[*]/div[*]):not(:has(img[src*="cdn.soft98.ir"])):not(:has-text(/)):not(:has(li)):not(:has(a[href*="/dl"])){ remove: true; }
 soft98.ir#$?#[href*="SMostafa.ir"] { remove: true; }
 ||soft98.ir/engine/modules/antibot/antibot.php
 ||soft98.ir/engine/ajax/controller.php
-||file.soft98.ir^$image,redirect=nooptext
-||cdn.hostdl.com^$image,redirect=nooptext
+||file.soft98.ir^$image,redirect=click2load.html
+||cdn.hostdl.com^$image,redirect=click2load.html
 ||linkdoni.soft98.ir^
 ||iranicard.ir^$domain=soft98.ir
+||iranicard.ir^$removeparam=/^utm_/
 ||asiatech.ir^$domain=soft98.ir
+||asiatech.ir^$removeparam=/^utm_/
 ||cdn.hostdl.com/goto/soft98
 ||sysban.ir^$domain=soft98.ir
 ||smostafa.ir^$domain=soft98.ir
--- a/EnglishFilter/sections/foreign.txt
+++ b/EnglishFilter/sections/foreign.txt
@@ -6452,10 +6452,10 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
 zoomit.ir#%#document.cookie = "BnShowNum=4; path=/;";
 ! soft98.ir - Persian Blocker rules (daily fix)
 !#if (!adguard_app_windows && !adguard_app_mac && !adguard_app_android && !adguard_ext_android_cb)
-soft98.ir#%#//scriptlet("remove-class", c6e1y2, *, stay)
+soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2", stay)
 soft98.ir#$?#[href*="linkdoni.soft98.ir"] { remove: true; }
-soft98.ir##[src*="file.soft98.ir"]{color: #437593 !important; background-color: #437593 !important; border: #437593 !important;}
-soft98.ir##[src*="cdn.hostdl.com"]{color: white !important; background-color: white !important; border: white !important;}
+soft98.ir#$#[src*="file.soft98.ir"]{color: #437593 !important; background-color: #437593 !important; border: #437593 !important;}
+soft98.ir#$#[src*="cdn.hostdl.com"]{color: white !important; background-color: white !important; border: white !important;}
 soft98.ir,~forum.soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="98"]){ remove: true; }
 soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/section[*]/header){ remove: true; }
 soft98.ir#$?#:xpath(/html/body/div[*]/div/div/section[*]/header){ remove: true; }
@@ -6480,9 +6480,7 @@ soft98.ir#$?#[href*="SMostafa.ir"] { remove: true; }
 ||smostafa.ir^$domain=soft98.ir
 !#endif
 !#if (adguard_app_windows || adguard_app_mac || adguard_app_android)
-soft98.ir##div[id^="c"][id$="header"]
-soft98.ir###si2debar > section.card:not([id])
-soft98.ir###sideb3r2_s1ticky
+! TODO: Add rules
 !#endif
 @@||soft98.ir^$generichide
 ! ||soft98.ir/templates/soft98_default/js/jquery.js$replace=/(\|download\|hcrack\|)[\s\S]*?(\|\|collapsing\|)/\$1body\$2/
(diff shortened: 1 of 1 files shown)


MasterKia:
@zloyden

zloyden (commented):
Is this update required? I don't see any difference with the new rules.

MasterKia:
Open one of the links in the site (Mozilla Firefox or Chrome for example) and scroll down, see if the red "سیسبان" text or the number "909....." at rhe end of the article is visible.

AdamWr:
It seems that some rules are incorrect.
These two:
```
soft98.ir##[src*="file.soft98.ir"]{color: #437593 !important; background-color: #437593 !important; border: #437593 !important;}
soft98.ir##[src*="cdn.hostdl.com"]{color: white !important; background-color: white !important; border: white !important;}
```
should be (missed `$`):
```
soft98.ir#$#[src*="file.soft98.ir"] { color: #437593 !important; background-color: #437593 !important; border: #437593 !important; }
soft98.ir#$#[src*="cdn.hostdl.com"] { color: white !important; background-color: white !important; border: white !important; }
```

This:
```
soft98.ir#%#//scriptlet("remove-class", c6e1y2, *, stay)
```
should be something like:
```
soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2")
```

And it looks like some elements (header) which are not ads are also removed (by a rule with `header`).

By the way, do you really need `xpath` in these rules?
Maybe some of them could be rewritten to not use `xpath`, if I'm not wrong, `xpath` is rather slow.

MasterKia:
This site changes CSS names too frequently to avoid filters.
For example it changes "linkdoni" to "li2nkdoni" or "link" to "lin2k".


As for the syntax compatibility, I'll fix them now and keep in mind for the next time (I'm accustomed to uBO syntax).

MasterKia:
@AdamWr I fixed everything you said.

EnglishFilter/sections/foreign.txt
```diff
@@ -6452,27 +6452,33 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
...
 ||asiatech.ir^$domain=soft98.ir
+||asiatech.ir^$removeparam=/^utm_/
 ||cdn.hostdl.com/goto/soft98
 ||sysban.ir^$domain=soft98.ir
 ||smostafa.ir^$domain=soft98.ir
 !#endif
 !#if (adguard_app_windows || adguard_app_mac || adguard_app_android)
-soft98.ir##div[id^="c"][id$="header"]
-soft98.ir###si2debar > section.card:not([id])
-soft98.ir###sideb3r2_s1ticky
```
zloyden, line 6475: No need to remove these rules.

EnglishFilter/sections/foreign.txt
```diff
@@ -6452,27 +6452,33 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
...
 soft98.ir#$?#[href*="SMostafa.ir"] { remove: true; }
-||soft98.ir/engine/modules/antibot/antibot.php
-||soft98.ir/engine/ajax/controller.php
-||file.soft98.ir^$image,redirect=nooptext
-||cdn.hostdl.com^$image,redirect=nooptext
+||file.soft98.ir^$image,redirect=click2load.html
+||cdn.hostdl.com^$image,redirect=click2load.html
 ||linkdoni.soft98.ir^
 ||iranicard.ir^$domain=soft98.ir
+||iranicard.ir^$removeparam=/^utm_/
```
zloyden, line 6473: We have AdGuard URL Tracking filter for such rules.

EnglishFilter/sections/foreign.txt
```diff
@@ -6452,27 +6452,33 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
 zoomit.ir#%#document.cookie = "BnShowNum=4; path=/;";
 ! soft98.ir - Persian Blocker rules (daily fix)
 !#if (!adguard_app_windows && !adguard_app_mac && !adguard_app_android && !adguard_ext_android_cb)
-soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="insta"]) { remove: true; }
+soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2", stay)
```
zloyden, line 6455: Is that `stay` necessary? If so, rewrite this as
```
soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2", "stay")
```

EnglishFilter/sections/foreign.txt
```diff
@@ -6452,27 +6452,33 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
...
 soft98.ir#$?#[href*="linkdoni.soft98.ir"] { remove: true; }
-soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/aside/section[*]/header) { remove: true; }
-soft98.ir#$?#:xpath(/html/body/div[*]/div[*]/aside/div/section/header) { remove: true; }
+soft98.ir#$#[src*="file.soft98.ir"]{color: #437593 !important; background-color: #437593 !important; border: #437593 !important;}
+soft98.ir#$#[src*="cdn.hostdl.com"]{color: white !important; background-color: white !important; border: white !important;}
```
zloyden, line 6458:
```
soft98.ir#$#[src*="file.soft98.ir"] { color: #437593 !important; background-color: #437593 !important; border: #437593 !important; }
```
Rewrite the second rule as I showed you above.

EnglishFilter/sections/foreign.txt
```diff
@@ -6452,27 +6452,33 @@ zoomit.ir#$#body > nav.is-sticky[style="top: 130px;"] { top: 0!important; }
...
+soft98.ir,~forum.soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="98"]){ remove: true; }
```
zloyden, line 6459: Add space before `{ remove: true; }`. Btw, if you want to exclude `forum.soft98.ir`, try `www.soft98.ir` instead of 2 domains.
```
soft98.ir,~forum.soft98.ir#$?#:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="98"]) { remove: true; }
```

zloyden (changes requested):

EnglishFilter/sections/foreign.txt
MasterKia: @AdamWr [said](https://github.com/AdguardTeam/AdguardFilters/pull/126697#issuecomment-1209277507): _And it looks like some elements (header) which are not ads are also removed (by a rule with header)._

AdamWr:
>uBO syntax.

It's AdGuard repo, so we would prefer to use AdGuard syntax. And as far as I know, it's also not exactly uBO syntax :)

Regarding header, I mean this rule:
```
soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/section[*]/header){ remove: true; }
```
It removes header which doesn't look like an ad.

image

</details>

>This site changes CSS names too frequently to avoid filters.

I'm aware of this and I mean that instead of something like this:
```
soft98.ir#$?#:xpath(/html/body/div[*]/div/div/section[*]/header){ remove: true; }
```
more or less something like this:
```
soft98.ir#$?#body > div > div > aside > section > header { remove: true; }
```
could be used (just example and as mentioned it removes hader which is not an ad).

<hr>

>I fixed everything you said.

This rather is still not valid:
```
soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2", stay)
```
It should be:
```
soft98.ir#%#//scriptlet("remove-class", "c6e1y2", ".c6e1y2", "stay")
```
https://github.com/AdguardTeam/Scriptlets/blob/master/wiki/about-scriptlets.md#remove-class

EnglishFilter/sections/foreign.txt
MasterKia: This insures that even if the ads isn't somehow blocked then the user isn't tracked.
If you click on the header and sidebar ads, they have tracking parameter.

If I was sure that the ads are gonna stay blocked then I wouldn't have used those removeparam rules.

MasterKia:
> It removes header which doesn't look like an ad.

Please suggest a rule that removes the "آسیاتک" header without relying on CSS names or "آسیاتک" text.
As with the CSS names, they do change them often.
As with the "آسیاتک" text, they tend to insert hidden unicode spaces in between the letters.

EnglishFilter/sections/foreign.txt
zloyden: Please check it with AdGuard URL Tracking filter. `utm` parameters should be removed.

EnglishFilter/sections/foreign.txt
MasterKia: Ok, added. But these depend on CSS names.

MasterKia:
> I'm aware of this and I mean that instead of something like this:
`soft98.ir#$?#:xpath(/html/body/div[*]/div/div/section[*]/header){ remove: true; }`
more or less something like this:
`soft98.ir#$?#body > div > div > aside > section > header { remove: true; }`
could be used (just example and as mentioned it removes hader which is not an ad).

That's a nice solution, but I've tried ":upward()" before and they added placebo `<span>` or `<div>` to neutralize my "upward"; so I suspect they'll do the same with your suggested filter but I think this doesn't affect xpath.

EnglishFilter/sections/foreign.txt
zloyden: I know, but they work fine in the app for now. Maybe we will come up with another solution for apps.

MasterKia:
I fixed the filters but GitHub doesn't let me edit (it's a bug that sometimes occur):
` The 'MasterKia/AdguardFilters' repository doesn't contain the 'EnglishFilter/sections/foreign.txt' path in 'MasterKia-patch-2/EnglishFilter/sections/foreign.txt'. `

zloyden:
How have you worked with our repo? It should be fine if you clone it to the MS VS Code or another IDE app.

AdamWr:
>Please suggest a rule that removes the "آسیاتک" header without relying on CSS names or "آسیاتک" text.

For example:
```
soft98.ir#$?#body > div > div > aside > section:first-of-type > header { remove: true; }
```

> That's a nice solution, but I've tried ":upward()" before and they added placebo nested `<span>` or `<div>` to neutralize my "upward"; `<div><div><div> ...`
> 
> So I suspect they'll do the same with your suggested filter but I think this doesn't affect xpath.

As far as I know, it doesn't matter if you use `xpath` or not, if they will add additional `div` or `span` then, for example rule like this:
```
soft98.ir#$?#:xpath(/html/body/div[*]/div/aside/section[*]/header){ remove: true; }
```
will stop working too.

By the way, I copied a wrong rule, so I updated my comment above.

MasterKia:
<details>
    <summary>Latest filters:</summary>

```
soft98.ir##+js(rc, c7e1y2, *, stay)
soft98.ir,~forum.soft98.ir##[src*="file.soft98.ir"]:style(color: #437593 !important; background-color: #437593 !important; border: #437593 !important;)
soft98.ir##:xpath(/html/body/header/div[*]/div/div) *:style(color: #437593 !important;)
soft98.ir##[src*="cdn.hostdl.com"]:style(color: white !important; background-color: white !important; border: white !important;)
soft98.ir##main + aside > section > div:style(color: white !important;)
soft98.ir,~forum.soft98.ir##:xpath(/html/body/div[*]/div/div[*]/div/div[1]):has(a[href*="98"]):remove()
soft98.ir##[href*="linkdoni.soft98.ir"]:remove()
soft98.ir##:xpath(/html/body/div[*]/div/aside/section[*]/header):remove()
soft98.ir##:xpath(/html/body/div[*]/div/div/section[*]/header):remove()
soft98.ir##:xpath(/html/body/div[*]/div/aside/section[*]/div):has(img[src*="cdn.hostdl.com"]):style(color: white !important;)
soft98.ir##:xpath(/html/body/div[*]/div/aside/div/section/header):remove()
soft98.ir##:xpath(/html/body/div[*]/div/aside/div/section/div):remove()
soft98.ir##:xpath(/html/body/div[*]/div/div/div/section):remove()
soft98.ir##:xpath(/html/body/div[*]/div[*]/main/article[*]/span):remove()
soft98.ir##:xpath(/html/body/div[*]/div[*]/main/article[*]/div[*]):not(:has(img[src*="cdn.soft98.ir"])):not(:has-text(/)):not(:has(li)):not(:has(a[href*="/dl"])):remove()
soft98.ir##[href*="SMostafa.ir"]:remove()
||file.soft98.ir^$image,redirect=click2load.html,domain=~forum.soft98.ir
||cdn.hostdl.com^$image,redirect=click2load.html
||linkdoni.soft98.ir^$all
||iranicard.ir^$all,domain=soft98.ir
||iranicard.ir^$removeparam=/^utm_/
||asiatech.ir^$all,domain=soft98.ir
||asiatech.ir^$removeparam=/^utm_/
||cdn.hostdl.com/goto/soft98^$all
||sysban.ir^$all,domain=soft98.ir
||smostafa.ir^$all,domain=soft98.ir
@@||soft98.ir^$ghide
```
</details>

zloyden:
> Given that sometimes I can't edit files and I have to try again half an hour later, what I can do for now is to send latest filters here and one of the maintainers updates the list.

That strange one. I'd recommend you to use VS Code, clone our repo to the app and do changes.

https://github.com/AdguardTeam/AdguardFilters/pull/126697#issuecomment-1209598073 thanks, but as I can see, those rules still work fine.

AdamWr:
I meant that in general Extended CSS rules are slower than simple cosmetic/element hiding rules, because the former need to run additional JavaScript code to hide elements.
So, instead of using something like:
```adblock
#?#:xpath(/html/body/div[*]/div/div/div/section)
```
something like this could be used:
```adblock
##html >body > div:not(:empty) > div > div > div > section
```
It should hides the same element(s) and should do this faster and it will require less resources.

By the way, you can use [selectors debug mode](https://github.com/AdguardTeam/ExtendedCss#selectors-debug-mode) or [ExtendedCss.query](https://github.com/AdguardTeam/ExtendedCss#debugging-extended-selectors) to check performance of the rules.

For example, for these rules:
```adblock
google.com#?#html > body > div > div > div > img
google.com#?#:xpath(/html/body/div/div/div/img)
```

Both rules are Extended CSS rules and if I understand correctly, `:xpath` rule required 2.9 ms, but not `:xpath` rule required 1.5 ms.

image

</details>

Anyway, in case like this, it would be better to use not Extended CSS rule:
```adblock
google.com##html > body > div > div > div > img
```

PR #127244, opened 2022-08-15
scopely.com

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/126099

### DIFF
--- a/MobileFilter/sections/specific_app.txt
+++ b/MobileFilter/sections/specific_app.txt
@@ -330,8 +330,6 @@
 !
 !
 !
-! https://github.com/AdguardTeam/AdguardFilters/issues/126099
-||adsgard.aprod.scopely.io^$app=com.pieyel.scrabble
 ! https://github.com/AdguardTeam/AdguardFilters/issues/125860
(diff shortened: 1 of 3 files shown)


MobileFilter/sections/specific_app.txt
```diff
@@ -22,6 +22,7 @@
 ||graph.facebook.com^$app=com.booking
 !
 ! Blocking Facebook mobile ads
+||graph.facebook.com^$app=com.pieyel.scrabble
```
Alex-302, line 25: This rule will break FB login("play with friends" button)

MobileFilter/sections/specific_app.txt
```diff
@@ -330,6 +331,8 @@
 !
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/126099
+||adsgard.aprod.scopely.io^$app=com.pieyel.scrabble
```
Alex-302, line 335: What about another games of this author?

Alex-302: And missed trecker:
https://489a19fa-d831-4954-858d-46436d5cdb20.event.collector.scopely.io

Alex-302 (changes requested):

Alex-302 (approved):

PR #127378, opened 2022-08-17
limehd.tv

scripthunter7:
https://github.com/AdguardTeam/AdguardFilters/issues/127220

### DIFF
--- a/EnglishFilter/sections/antiadblock.txt
+++ b/EnglishFilter/sections/antiadblock.txt
@@ -11169,3 +11169,6 @@ wolframclient.net#@#.adsbygoogle
 bigbrothercanada.ca#%#//scriptlet('set-constant', 'video_player_config.players.0.adblock', 'false')
 !+ PLATFORM(ios, ext_android_cb)
 @@||bigbrothercanada.ca^$generichide
+! https://github.com/AdguardTeam/AdguardFilters/issues/127220
+limehd.tv#%#//scriptlet('set-constant', 'window.ya', 'true')
+limehd.tv#%#//scriptlet('set-constant', 'window.google', 'true')


EnglishFilter/sections/antiadblock.txt
```diff
@@ -11169,3 +11169,6 @@ wolframclient.net#@#.adsbygoogle
 bigbrothercanada.ca#%#//scriptlet('set-constant', 'video_player_config.players.0.adblock', 'false')
 !+ PLATFORM(ios, ext_android_cb)
 @@||bigbrothercanada.ca^$generichide
+! https://github.com/AdguardTeam/AdguardFilters/issues/127220
+limehd.tv#%#//scriptlet('set-constant', 'window.ya', 'true')
+limehd.tv#%#//scriptlet('set-constant', 'window.google', 'true')
```
zloyden, line 11174: @scripthunter7 there are some points that I want to show you.
1. This is a Russian website, so the rules have to be added to AdGuard Russian (not at the end of the file).
2. Your rules are good, but the control panel does not appear.

image

</details><br/>

Fixed directly https://github.com/AdguardTeam/AdguardFilters/commit/e5df625dcff9fa4a005cf48a73505d2b39bd74b1

scripthunter7: @zloyden 

> This is a Russian website, so the rules have to be added to AdGuard Russian

I was just looking at TLD, but you are right.

> but the control panel does not appear

You are right again. At the beginning, the deactivated control appears for a short time (and I should have tested more thoroughly here), but the script makes the control available only after the ads. However, without the IMA3 API, the `adsInitialization` function will fail at some point. The script hides the controller using a class, and I didn't removed this class:
```
limehd.tv#%#//scriptlet('remove-class', 'video-sdk-container--up', '#video-sdk-container', 'stay')
```

Although your solution is more optimal, since it is only an exception.

Thanks for the tips and sorry! 🙂

zloyden: I had checked the site and decided to add your rules with directives.
Thanks and you are welcome to create new pull requests:)

PR #128408, opened 2022-08-30
Disabling overwriting clipboard

BlazDT:
More details here: https://bugs.chromium.org/p/chromium/issues/detail?id=1334203
Test page: https://output.jsbin.com/yuliwoz/quiet

Feature can be used to put advertising in clipboard.


zloyden (approved):

AdamWr:
I suppose that it also can be used to useful things (like, for example, button to copy image/text to clipboard, or something similar), maybe it would be better to add it to `Tracking Protection` filter, though I'm still not sure.
By the way, it seems that there is also `navigator.clipboard.writeText` which can also be abused.

BlazDT:
I added to main filter as it can be abused for every purpose.

Alex-302:
Many users use password managers. How will this protect them?

BlazDT:
Added other method to write and both .read (reading rules should be removed when Chrome fixed the issue).

krystian3w:
Demo no longer work in Canary 107.

BlazDT:
So seems to be fixed by Google :)

krystian3w:
Or Chromium community or MS team (this is not related with Alphabet corporation) and planned publish on 25 October.

PR #130114, opened 2022-09-19
10play.com.au

jellizaveta:
https://github.com/AdguardTeam/AdguardFilters/issues/130107

Requests to log in when the account is already logged in
After removing rule the anti-adblock does not appear

Screenshot 2022-09-19 at 10 58 55

</details><br/>


jellizaveta:
1. Visit to the site from an Australian IP and register
2. After registering leave only the annoyance filter 
3. Then go to https://10play.com.au/live

Alex-302 (approved):

PR #132240, opened 2022-10-15
Remove obsoleted rule

scripthunter7:


SocialFilter/sections/general_extensions.txt
```diff
@@ -232,7 +232,7 @@ crimea-kurort.com#$#img[width="88"][height="31"] {visibility: hidden!important;}
 maskme.ru#$#img[width="88"][height="31"] {visibility: hidden!important;}
 !
 gamesradar.com#$#.shareSocialArticle-wrap { visibility: hidden; }
-softwares-serials.in#$##FTslide { visibility: hidden;!important; }
+softwares-serials.in#$##FTslide { visibility: hidden!important; }
```
zloyden, line 235: We also add space after the value.
```
softwares-serials.in#$##FTslide { visibility: hidden !important; }
```

Btw, is that domain alive?

zloyden (changes requested):

SocialFilter/sections/general_extensions.txt
scripthunter7: @zloyden As I see, no :sweat_smile: So I remove the rule instead

scripthunter7: I'm just testing the parser

scripthunter7: Done

zloyden (approved):

PR #133098, opened 2022-10-25
www2.myjcom.jp

spiri-leo:
Issue: #133069


SocialFilter/sections/specific.txt
```diff
@@ -57,6 +57,7 @@ polaris-official.com,apexheadline.com,aucfan.com,sist8.com,spaceshipearth.jp,sou
...
 !
+myjcom.jp##.sp_sns
```
zloyden, line 60: It does not remove a share on desktop. Fixed directly.

PR #133876, opened 2022-11-02
doujinhibiki.net

tachibana-shin:
# Creating the pull request

> It is highly recommended to use our [Web Reporting Tool](https://kb.adguard.com/en/technical-support/reporting-tool) instead of creating an issue on GitHub directly\

- [ ] Filters maintenance.

### Enter the issue address
https://github.com/AdguardTeam/AdguardFilters/issues/133617

0. **DO NOT** upload screenshots with sexually explicit material on GitHub directly.\
 Instead, upload it to a third-party image hosting and post the link to it without preview (!) here.\
 Also, mention if the link leads to NSFW content;

1. Add screenshots of the problem. You can drag and drop images or paste them from clipboard

    Use `<details> </details>` tag to hide screenshots under the spoiler;

2. Describe the issue in detail until it is absolutely clear from the screenshot what the problem is

    You can also indicate any other information that you think the developers should know.

**Warning:** Please remove personal information before uploading screenshots!

1. Your comment

2. Screenshots

<details>

<summary>Screenshot 1:</summary>

</details>

<details>

<summary>Screenshot N:</summary>

</details>

### Terms

### DIFF
--- a/JapaneseFilter/sections/general_elemhide.txt
+++ b/JapaneseFilter/sections/general_elemhide.txt
@@ -62,6 +62,8 @@
 ##.ad_textlink_box
 ##.ad_thumbnail_header
 ##.adbox
+! https://github.com/AdguardTeam/AdguardFilters/issues/133617
+##.adBox
 ##.adgoogle
 ~hgh14x.com##.adlist
 ##.admax-ads
--- a/JapaneseFilter/sections/general_elemhide.txt
+++ b/JapaneseFilter/sections/general_elemhide.txt
@@ -62,8 +62,6 @@
 ##.ad_textlink_box
 ##.ad_thumbnail_header
 ##.adbox
-! https://github.com/AdguardTeam/AdguardFilters/issues/133617
-##.adBox
 ##.adgoogle
 ~hgh14x.com##.adlist
 ##.admax-ads
--- a/JapaneseFilter/sections/specific.txt
+++ b/JapaneseFilter/sections/specific.txt
@@ -4663,3 +4663,5 @@ gtoys.blog.fc2.com##.adsbygoogle
 gtoys.blog.fc2.com###adleft
 !+ PLATFORM(ext_ff, ext_opera, ios, ext_android_cb, ext_ublock)
 gtoys.blog.fc2.com###adright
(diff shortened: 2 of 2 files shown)


JapaneseFilter/sections/general_elemhide.txt
```diff
@@ -62,6 +62,8 @@
 ##.ad_textlink_box
 ##.ad_thumbnail_header
 ##.adbox
+! https://github.com/AdguardTeam/AdguardFilters/issues/133617
+##.adBox
```
zloyden, line 66: Why do you add this rule as generic?
It should be something like that and you need to add it to the specific.txt.
```
doujinhibiki.net##div[class^="unBox"][class*="ad"]
```

zloyden (changes requested):

JapaneseFilter/sections/general_elemhide.txt
tachibana-shin: > Why do you add this rule as generic? It should be something like that and you need to add it to the specific.txt.
> 
> ```
> doujinhibiki.net##div[class^="unBox"][class*="ad"]
> ```

this rule is just a variation of the `.adbox` rule that is used a lot in Japanese websites. so i think instead of defining a website we should put it in generic

zloyden: > ##.adBox

Better to do it in separate pull request with examples. Btw, this rule does not block leftovers mentioned in the task.

zloyden: I fixed this directly and don't see the need to add this rule to the current site.
@Yuki2718 what do you think about adding the rule above as generic one? Is that really needed?

zloyden: Anyway, we have `##.adBox` in AdGuard Base.

Yuki2718: As you said `##.adBox` is in optimzed EL so I don't understand how it was an issue - I don't see $generichide even in content_blocker.txt. Could you reproduce?

zloyden: Nope, I can't reproduce it. Then it's ok, thanks for checking this out :)

PR #134630, opened 2022-11-11
Fix regex for uBO (addresses https://github.com/DandelionSprout/adfil…

Yuki2718:
…t/discussions/163#discussioncomment-4117315)

# Creating the pull request

> It is highly recommended to use our [Web Reporting Tool](https://kb.adguard.com/en/technical-support/reporting-tool) instead of creating an issue on GitHub directly\

- [x] Filters maintenance.

### Enter the issue address

https://github.com/DandelionSprout/adfilt/discussions/163#discussioncomment-4117315
It turned out uBO requires `-` to be escaped if it's at the end.

### Terms


Alex-302:
Our regexes are valid. The most of filters have rules with `-]` in regexes.

Yuki2718:
Then how about `NOT_PLATFORM(ext_ublock)`? uBF has counterparts of them anyway.

scripthunter7:
I think that this issue can also occur in the case of many filter lists, since `-]` is a valid regex, so this PR does not solve the problem itself. Maybe a better solution would be to replace `-]` with `\-]` internally before the regexp is passed to the [analyzer](https://github.com/foo123/RegexAnalyzer/blob/master/src/js/Regex.js).

Yuki2718:
So move to https://github.com/AdguardTeam/FiltersCompiler ?

scripthunter7:
> So move to https://github.com/AdguardTeam/FiltersCompiler ?

I think this issue is related to the uBlock parser:
https://github.com/gorhill/uBlock/blob/master/src/js/static-filtering-parser.js#L26

I guess uBO also uses lists that don't go through the AG's Filters Compiler, and those lists can easily use `-]` in regexps.

Yuki2718:
@gorhill

scripthunter7:
By the way, here is the problematic part in the analyzer library:

https://github.com/foo123/RegexAnalyzer/blob/807f106914ed8b650d792d7a158f63040fd579c3/src/js/Regex.js#L1079-L1099

When line 1094 checks whether the currently iterated character is a hyphen, you need a lookahead in which you check whether the next character sequence is whitespaces + closing bracket.

But this seems like a rather outdated / low documented / low tested library, I think there are much more modern ones.

scripthunter7:
Without any warranty, something like this: https://github.com/scripthunter7/RegexAnalyzer/blob/master/src/js/Regex.js

gorhill:
We discussed this a long time ago, I said I would fix this eventually, we should bother AdGuard's repo just because I take long to fix stuff which is one thing among all the rest I need to work on.

danny0838:
@gorhill About the [commit](https://github.com/gorhill/uBlock/commit/d51b7e082b73db745a1f8321e65b086902d80d80#diff-8a8c24285f93dbeb182e3383fd0907004bcd92ae6e2009fcb9c84e1dd5877b7c) (I cannot leave a comment in the commit page, seems to be an issue of GitHub?):

I wonder if modifying the regex.js library without populating to the upstream is a good idea. Also this doesn't seem to deal with the `[-abc]` case.

I would recommend preprocessing regex strings by escaping every `-` with `\-` (exclude already escaped ones like `\-`, of course) before sending to the analyzer instead.

krystian3w:
>  (I cannot leave a comment in the commit page, seems to be an issue of GitHub?)

That is limitation to disable open issues in code of repository.

Sometimes someone can turn it off - once I managed to give a reaction to PR and after a few days someone blocked it undo/new.

PR #139662, opened 2023-01-09
pcgames.de - cookie-auto-accept

BlazDT:
If cookies are not accepted, the quizzes are not showing up.

https://www.pcgames.de/Harry-Potter-Brands-23301/Specials/Quiz-Harry-Potter-Filme-Buecher-Hogwarts-1410748/


zloyden (approved):
Should we add such rules via the pull request?
Anyway, it works fine on my end.

BlazDT:
We discussed in chat that more difficult JS rules are to be added via pull request. Since it uses mutationobserver it is imho a more difficult JS rule.

zloyden:
I'd read this. Ok, I got it.

Issue #139027, opened 2023-01-03
Replace `#%#Object.defineProperty` rules by scriptlets


Alex-302:
### Prerequisites

- [X] I checked the documentation and found no answer;
- [X] I checked to make sure that this issue has not already been filed;
- [X] This is not an ad/bug report.

### Problem description

The most of `#%#Object.defineProperty` rules can by replaced by `set-constant` scriptlet.

### Proposed solution

.

DandelionSprout:
I approve of this project since uBlock Origin supports scriptlets, but not `#%#` (from what I'm aware of).

Alex-302:
@AdamWr @jellizaveta @piquark6046 *ping*
Please finish the last 20 rules.

piquark6046:
Some JS rules returning a value (`Object.defineProperties(window, { _wm: { get: function(){ return null; }}})`) can be converted to scriptlet when scriptlet 1.9.37 is widespread.

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/139027#issuecomment-1611408188
@piquark6046 @jellizaveta Some rules like that are very old, and maybe unnecessary. You can check 5-10 domains of each rule and search in resources. 
For example `#%#Object.defineProperties(window, { "_impspcabe_alpha": { value: false, writable: false }, "_impspcabe_beta": { value: false, writable: false }, "_impspcabe_path": { value: 'about:blank', writable: false }, "_impspcabe": { value: false, writable: false } });` - search in resources for `_impspcabe_alpha` (better in Fiddler - there one search can be applied for a bunch of sites)

yuhanchn01:
@Alex-302 one more rule left

![chn22](https://github.com/AdguardTeam/AdguardFilters/assets/151660568/00cd7be0-49fe-4260-b822-9fea20edd73c)

PR #139872, opened 2023-01-12
Obsolete rules; JS => scriptlet (#139872) 

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/139027


BaseFilter/sections/general_extensions.txt
```diff
@@ -2154,7 +2154,7 @@ mr-jatt.in#%#//scriptlet("json-prune", "rot_url pop_type")
 mr-jatt.in#%#//scriptlet("prevent-addEventListener", "", "x20click")
 ! https://github.com/AdguardTeam/AdguardFilters/issues/59478
 ! popunder when click on the back button
-vrsumo.com#%#Object.defineProperty(window, 'offer', { value: ''});
+vrsumo.com#%#//scriptlet('set-constant', 'offer', '')
```
zloyden, line 2157: I'll revert this and open an issue in the scriptlets.
When you open this link and click on the back button on mouse, you get redirected. The scriptlet does not work, but the js rule does.

Alex-302: Hmm. I checked it and was no redirects.

zloyden: I've found another scriptlet that works fine, replaced the js rule.

PR #140499, opened 2023-01-19
remove bytedance related domain, block this domain may cause many byt…

yangzeen2008:
…edance related APP malfunction, such as Tiktok, Pico, Capcut and so on

When I use adguard and use this filter, I found i cannot make payment in PICO APP and tiktok APP. So I checked the blocked domain and found this.
So I removed this bytedance related domain, block this domain may cause many bytedance related APP malfunction, such as Tiktok, Pico, Capcut and so on.

### DIFF
--- a/MobileFilter/sections/adservers.txt
+++ b/MobileFilter/sections/adservers.txt
@@ -216,7 +216,6 @@
 ||static.hyprmx.com^
 ||prod.cm.publishers.advertising.a2z.com^
 ||events-*.op.hicloud.com^
-||isnssdk.com^
 ||app.adjust.world^
 ||app.adjust.net.in^
 ||ads.feelingtouch.com^


Alex-302:
Replaced the rule by AdAway's. Please check after filters update.

jordanswrld94:
Can we remove log*-applog-* tiktokv.com from the adguard tracking protection list. It breaks tiktok when you first install it. You cant login or anything til its unblocked. @Alex-302

Alex-302:
Can't confirm - installed and registered using google account.

jordanswrld94:
Im guessing its because your in a different country? That tracker is us based @Alex-302 also can we block log-va.tiktokv.com i tested throughly and nothing is broken

jordanswrld94:
Also. i.inssdk.com breaks tikok thumbnails bad filter @Alex-302

Alex-302:
We do not block i.inssdk.com

PR #142392, opened 2023-02-09
glasgowlive.co.uk

zloyden:
https://github.com/AdguardTeam/AdguardFilters/issues/142387
```javascript
AG_onLoad(function() {
    if (!document.cookie.includes('euconsent-v2')) {
        var g = new MutationObserver(function() {
            var el = document.querySelector('.qc-cmp2-footer button[mode="primary"]');
            el && el.click();
        });
        g.observe(document, {
            childList: !0,
            subtree: !0
        });
        setTimeout(function() {
            g.disconnect()
        }, 1E4)
    }
});
```


AdamWr:
It looks like similar to this case - https://github.com/AdguardTeam/AdguardFilters/issues/138817
So this rule:
```adblock
glasgowlive.co.uk#%#//scriptlet('set-constant', 'chameleonVideo.adDisabledRequested', 'true')
```
should fix video player and it shouldn't be necessary to unblock cookie notice and other requests.

zloyden:
It works fine, thanks! I've done necessary changes https://github.com/AdguardTeam/AdguardFilters/commit/902f011bd2ab33f8dda363efb0b75b5ae1dc45c2.

PR #142529, opened 2023-02-10
Integrate AGLint

scripthunter7:
Integration of the first version of AGLint to validate adblock rules. Currently, there aren't too many linter rules (yet). Therefore, in the first round, mainly syntax errors are filtered.

@ameshkov I would prefer to add the Husky pre-commit hook in another PR (if needed).


scripthunter7:
@ameshkov @Alex-302 The status of the linter for the master branch can also be displayed in a badge, please see the following link:
https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/adding-a-workflow-status-badge#using-the-branch-parameter

The link would be something like this:
https://github.com/AdguardTeam/AdguardFilters/actions/workflows/aglint.yml/badge.svg?branch=master

ameshkov:
@scripthunter7 great idea, let's add it

scripthunter7:
@ameshkov Added, but please note that the workflow badge will only appear after the first run

ameshkov (changes requested):
We should also add AGLint configuration file so that people could understand what rules there are

ameshkov (approved):

PR #166018, opened 2023-11-12
add new parameters for TikTok

fenopa:
refs:
https://github.com/Smile4ever/Neat-URL/issues/235
https://github.com/ThatRedKite/thatkitebot/blob/master/data/detrackparams.toml
https://github.com/mtxadmin/ublock/blob/master/filters/_removeparams
https://github.com/DrKain/tidy-url/blob/main/data/rules.js
https://github.com/DandelionSprout/adfilt/blob/master/LegitimateURLShortener.txt

### DIFF
--- a/TrackParamFilter/sections/specific.txt
+++ b/TrackParamFilter/sections/specific.txt
@@ -990,6 +990,21 @@ $removeparam=scid,domain=rakuten.co.jp|rakuten-wallet.co.jp
 ||tiktok.com^$removeparam=share_app_id
 ||tiktok.com^$removeparam=share_author_id
 ||tiktok.com^$removeparam=tt_from
+! TikTok
+||tiktok.com^$removeparam=_d
+||tiktok.com^$removeparam=checksum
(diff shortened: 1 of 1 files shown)


Alex-302:
Need example values and links with these params.

fenopa:
there are some examples in ref links

fenopa:
@BlazDT @jellizaveta can someone take a look? I spent time to compile these and it's getting closed without any respect.

BlazDT:
Alex meant link to every tracking parameter with vues and how to reproduce them.
Some may sound like a tracker, but provide some function on the page.

fenopa:
> Some may sound like a tracker, but provide some function on the page.

@BlazDT No they do not, all of these ref links for "tracking parameters", why would they include parameter that provide function to remove as tracking. The currently existing were added with a DandelionSprout/adfilt discussion link, this only updates them with that and some other links.

Alex-302:
I don't see params in urls. Need real actual examples, and description where we can get them.

PR #166387, opened 2023-11-16
update eksisozluk1923.com

fenopa:
# Creating the pull request

- [x] Filters maintenance.

### Terms


Alex-302:
Hi
I don't see these elements.

fenopa:
> Hi I don't see these elements.
Hi, it looks like there's no "search box ad" at this moment (there was when I submitted this PR) but it's an ad type they do frequently, so it would be nice to add this filter.

fenopa:
@Alex-302 there is an ad with this type right now.

fenopa:
@Alex-302 there is one right now as well.

fenopa:
@zloyden @BlazDT can you merge this?

BlazDT:
Are there any safer selectors than the div style color one? Any other children?
Also when I compare both changes only different colour.

fenopa:
> Are there any safer selectors than the div style color one? Any other children? Also when I compare both changes only different colour.

@BlazDT  I don't know, I just used the existing filter for a previous version of the same ad and modified the color according to the new one. Can you come up with a better filter to remove that item? (you can type something random in search box, and the first item in the autocomplete box would be the one that should be removed)

TurkishFilter/sections/specific.txt
```diff
@@ -222,7 +222,10 @@ manga-tr.com#?#.col-sm-4 > .well3:has(> .panel > .panel-heading2 > center:contai
 eksisozluk1923.com#$?##interstitial-close-link-source { remove: true; }
 eksisozluk1923.com#$?#.ad-doubleclickwebinterstital { remove: true; }
 eksisozluk1923.com#$?#.ui-autocomplete > li:has(> div[style="color:rgb(0, 0, 0)"]) { remove: true; }
+eksisozluk1923.com#$?#.ui-autocomplete > li:has(> div[style="color:rgb(8, 8, 8)"]) { remove: true; }
```
zloyden, line 225: I guess this rule should be enough instead of similar ones.
```adblock
eksisozluk1923.com#$?#.ui-autocomplete > li:has(> div[style^="color:rgb("]) { remove: true; }
```

fenopa: @zloyden modified it according to your suggestion. tested it and it looks good.

TurkishFilter/sections/specific.txt
```diff
@@ -221,8 +221,7 @@ manga-tr.com#?#.col-sm-4 > .well3:has(> .panel > .panel-heading2 > center:contai
 [$domain=lifebursa.com,path=/^\/$/]#?#.container > div.row:has(> div[class^="col-"] > div.banner)
 eksisozluk1923.com#$?##interstitial-close-link-source { remove: true; }
 eksisozluk1923.com#$?#.ad-doubleclickwebinterstital { remove: true; }
-eksisozluk1923.com#$?#.ui-autocomplete > li:has(> div[style="color:rgb(0, 0, 0)"]) { remove: true; }
-eksisozluk1923.com#$?#.ui-autocomplete > li[style^="position: absolute;"][style*="z-index:"]:has(> .ui-menu-item-wrapper:contains(/^sponsorlu$/)) { remove: true; }
+eksisozluk1923.com###search-box-ad
```
zloyden, line 224: Why the rules were removed? It's not enough on my end. I'll add the suggested one.

zloyden (changes requested):

zloyden (approved):

PR #170814, opened 2024-01-15
Changed descriptions of sections

Alex-302:
**Purpose:** 
We will add new non-antiadblock CSS, JS/scriptlets, HTML filtering/`$replace`-like rules to `specific.txt` files.
Existing rules will be moved to `specific.txt` gradually to prevent file structure disruption and redundant commenting/grouping.


BaseFilter/sections/specific.txt
```diff
@@ -2,7 +2,7 @@
 ! This section contains the list of domain-specific rules that block ads.
 !
 ! Good: example.org###rek; ||example.com/ads/
-! Bad: example.org#$##rek { display: none !important; } (for instance, should be in general_extensions.txt)
+! Bad: example.org#@##banner_ad (for instance, should be in general_extensions.txt or in antiadblock.txt)
```
zloyden, line 5: > We will add new non-antiadblock CSS, JS/scriptlets, HTML filtering/$replace-like rules to specific.txt files.

Then why do we need `general_extensions.txt`?

> (for instance, should be in general_extensions.txt or in antiadblock.txt)

Perhaps it should be `allowlist.txt` instead of `general_extensions.txt`. But I don't recall when we add such exceptions in `allowlist.txt`. Seems to be a rare case.

Alex-302: > Then why do we need general_extensions.txt?

We will remove them after moving rules to new place.

> Perhaps it should be allowlist.txt instead of general_extensions.txt.

yes, wrong copy-pasta

zloyden (approved):

AdamWr (approved):

PR #170936, opened 2024-01-17
Update yt

Yuki2718:

- [x] AdGuard gets detected on a website;
- [ ] Filters maintenance.

## What issue is being fixed?

https://github.com/AdguardTeam/AdguardFilters/issues/170786

Details are on Slack. As the anti-adb is only sent in www, limited to this.

### Terms

### DIFF
--- a/BaseFilter/sections/specific.txt
+++ b/BaseFilter/sections/specific.txt
@@ -154,6 +154,13 @@ youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', '
 youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', 'ytInitialPlayerResponse.playerAds', 'undefined')
 youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', 'playerResponse.adPlacements', 'undefined')
 !
+! Remove anti-adb
+!#if (adguard_app_windows || adguard_app_mac || adguard_app_android || adguard_ext_firefox)
+.com/watch?v=$replace=/"auxiliaryUi.*?e\}\}\}\,//,domain=www.youtube.com
+.com/watch?v=$replace=/("trackingParam":"kx_fmPxhoPZR)[-_0-9A-Za-z]{275}([-_0-9A-Za-z]{34}zRgkuMsNLBwOcCE59TDtslLKPQ-SS"\})/\$1\$2/,domain=www.youtube.com
+/youtubei/v*/player?key=$replace=/"auxiliaryUi.*?e\}\}\}\,//,domain=www.youtube.com
+/youtubei/v*/player?key=$replace=/("trackingParam":"kx_fmPxhoPZR)[-_0-9A-Za-z]{275}([-_0-9A-Za-z]{34}zRgkuMsNLBwOcCE59TDtslLKPQ-SS"\})/\$1\$2/,domain=www.youtube.com
+!#endif
 ! Modify YT xhr responses
 .com/watch?v=$xmlhttprequest,jsonprune=\$..[adPlacements\, adSlots\, playerAds],domain=youtubekids.com|youtube-nocookie.com|youtube.com
 /youtubei/v*/player?key=$jsonprune=\$..[adPlacements\, adSlots\, playerAds],domain=youtubekids.com|youtube-nocookie.com|youtube.com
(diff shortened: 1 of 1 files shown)


BaseFilter/sections/specific.txt
```diff
@@ -154,6 +154,13 @@ youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', '
 youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', 'ytInitialPlayerResponse.playerAds', 'undefined')
 youtubekids.com,youtube-nocookie.com,youtube.com#%#//scriptlet('set-constant', 'playerResponse.adPlacements', 'undefined')
 !
+! Remove anti-adb
+!#if (adguard_app_windows || adguard_app_mac || adguard_app_android || adguard_ext_firefox)
+.com/watch?v=$replace=/"auxiliaryUi.*?e\}\}\}\,//,domain=www.youtube.com
+.com/watch?v=$replace=/("trackingParam":"kx_fmPxhoPZR)[-_0-9A-Za-z]{275}([-_0-9A-Za-z]{34}zRgkuMsNLBwOcCE59TDtslLKPQ-SS"\})/\$1\$2/,domain=www.youtube.com
+/youtubei/v*/player?key=$replace=/"auxiliaryUi.*?e\}\}\}\,//,domain=www.youtube.com
+/youtubei/v*/player?key=$replace=/("trackingParam":"kx_fmPxhoPZR)[-_0-9A-Za-z]{275}([-_0-9A-Za-z]{34}zRgkuMsNLBwOcCE59TDtslLKPQ-SS"\})/\$1\$2/,domain=www.youtube.com
```
Alex-302, line 162: Rules with `zRgkuMsNLBwOcCE59TDtslLKPQ` don't match on my end
https://regex101.com/r/gKZTp6/1

Yuki2718: Okay, will change to `[0-9A-Za-z]{9}BwOcCE59TDtslLKPQ`

Alex-302:
@AdamWr @piquark6046 Could you please take a look? 
On my end rules applied, but without them end disabled scriptlets I can't trigger anti-adblock.

AdamWr:
If I understand correctly, these rules are applied only if there is adblock detection (a real detection, not invoked by a special script or something like that).
If so, then I cannot check if it works or not, because there is no detection on my account, but I don't see any breakage with these rules.

Alex-302:
Let's wait few days and check if the rules still match that parts.

Yuki2718:
> Let's wait few days and check if the rules still match that parts.

Already a week passed, still matches. Their encoding doesn't change until they discover this and counter, only concern is the encoded text itself.
@AdamWr @piquark6046 Are `kx_fmPxhoPZR` and `BwOcCE59TDtslLKPQ-SS` present in your `responseContext.mainAppWebResponseContext.trackingParam`?

Yuki2718:
Tested on the Windows app, works perfectly against real anti-adb.

piquark6046:
https://github.com/AdguardTeam/AdguardFilters/pull/170936#issuecomment-1898171743
`responseContext.mainAppWebResponseContext.trackingParam` does not exist in my personal Google account.

Yuki2718:
@piquark6046 You have to click some videos, then check `player?key=` requests in Network tab of devtools.

AdamWr:
> Are `kx_fmPxhoPZR` and `BwOcCE59TDtslLKPQ-SS` present in your `responseContext.mainAppWebResponseContext.trackingParam`?

Yes, I see it, but it looks like that rules are not applied on my end, so I guess that this `trackingParam` is different when detections occur.

piquark6046:
> Are kx_fmPxhoPZR and BwOcCE59TDtslLKPQ-SS present in your responseContext.mainAppWebResponseContext.trackingParam?

OK. I found `kx_fmPxhoPZR` and `BwOcCE59TDtslLKPQ-SS` without any account.

Yuki2718:
> Yes, I see it, but it looks like that rules are not applied on my end, so I guess that this trackingParam is different when detections occur.

You mean not applied as the rule targets longer strings? That's expected - they add strings about anti-adb to `trackingParam` if they send anti-adb. The original `trackingParam` is still preserved, but interesingly it is in the tail dispite in the head in decoded text.

AdamWr:
>You mean not applied as the rule targets longer strings?

Yes, that's what I meant.

Yuki2718:
I found shorter strings in requests saved three monts ago. Will adjust the length. Note, as already mentioned, the rules won't apply usually - they match only if the warning is sent. If any problem arises, it must be only when the case otherwise user see anti-adb and we can tell such reports from characteristic error in the player.

PR #171886, opened 2024-01-31
Husky v9

scripthunter7:
Update Husky to v9.

See migration guide: https://github.com/typicode/husky/releases/tag/v9.0.1

### DIFF
--- a/.husky/install.mjs
+++ b/.husky/install.mjs
@@ -1,0 +1,11 @@
+// See: https://typicode.github.io/husky/how-to.html#ci-server-and-docker
+
+// Do not initialize Husky in CI environments. GitHub Actions set the CI env variable automatically.
+// See: https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables
+if (process.env.CI === 'true') {
+    process.exit(0);
+}
+
+// Initialize Husky programmatically.
+const husky = (await import('husky')).default;
+console.log(husky());
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,3 +1,8 @@
+#!/usr/bin/env sh
+. "$(dirname -- "$0")/_/husky.sh"
+
+# TODO: Remove first 2 lines: https://github.com/AdguardTeam/AdguardFilters/pull/171886#issuecomment-1918731193
+
 # See https://git-scm.com/docs/githooks#_post_merge
 
 # This function checks if a file has changed between the current HEAD and the previous HEAD
--- a/.husky/post-merge
+++ b/.husky/post-merge
@@ -1,6 +1,3 @@
-#!/usr/bin/env sh
-. "$(dirname -- "$0")/_/husky.sh"
-
 # See https://git-scm.com/docs/githooks#_post_merge
 
 # This function checks if a file has changed between the current HEAD and the previous HEAD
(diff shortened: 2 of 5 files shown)


ameshkov (approved):

scripthunter7:
@ameshkov @Alex-302 There is an issue that you should know about. According to [the migration guide](https://github.com/typicode/husky/releases/tag/v9.0.1), Husky 9 no longer requires the presence of the following 2 lines in the hook scripts:

```sh
#!/usr/bin/env sh
  . "$(dirname -- "$0")/_/husky.sh"
```

The problem: if a dependency is updated in the GitHub repo, we have to update it in the local repo as well. This is currently automatic in the filters repo and is done through Husky's `post-merge` hook. So if the user pulls the changes and the `package.json` or `yarn.lock` changed between the old HEAD and the new HEAD, `yarn install --force` is automatically called. This is a convenient behaviour in this repo and helps keep the linter results consistent, no one forgets to update.

However, since Husky 9 no longer requires these 2 lines, and I removed them in this PR, the following will happen, if we merge this PR in its current form:
- the user currently has Husky 8 installed locally
- the user pulls the latest changes from GitHub, so the 2 lines disappear from their local `post-merge`
- the locally installed Husky 8 tries to run the new `post-merge` hook but won't be able to spawn because the shebang is missing

So there are two options:
- the user manually calls the "yarn install --force" command in this one case (then Husky 9 will be installed locally, the automatic process will work again)
- we keep the lines - in theory, Husky 9 is backwards compatible, although I don't know if it will ever be marked as deprecated, currently `_/husky.sh` is just an empty file in Husky 9, we will probably have to make this change at some point anyway

Andrey, what do you suggest?

scripthunter7:
@ameshkov @Alex-302 There is an issue that you should know about. According to [the migration guide](https://github.com/typicode/husky/releases/tag/v9.0.1), Husky 9 no longer requires the presence of the following 2 lines in the hook scripts:

```sh
#!/usr/bin/env sh
  . "$(dirname -- "$0")/_/husky.sh"
```

The problem: if a dependency is updated in the GitHub repo, we have to update it in the local repo as well. This is currently automatic in the filters repo and is done through Husky's `post-merge` hook. So if the user pulls the changes and the `package.json` or `yarn.lock` changed between the old HEAD and the new HEAD, `yarn install --force` is automatically called. This is a convenient behaviour in this repo and helps keep the linter results consistent, no one forgets to update.

However, since Husky 9 no longer requires these 2 lines, and I removed them in this PR, the following will happen, if we merge this PR in its current form:
- the user currently has Husky 8 installed locally
- the user pulls the latest changes from GitHub, so the 2 lines disappear from their local `post-merge`
- the locally installed Husky 8 tries to run the new `post-merge` hook but won't be able to spawn because the shebang is missing

Also, `pre-commit` hook will fail for the same reason.

So there are two options:
1. the user manually calls the `yarn install --force` command in this case (then Husky 9 will be installed locally, the automatic process will work again) - no need to call it again manually after that
2. we keep the lines - in theory, Husky 9 is backwards compatible, although I don't know if it will ever be marked as deprecated, currently `.husky/_/husky.sh` is just an empty file in Husky 9 to avoid errors, we will probably have to make this change at some point anyway

What do you suggest?

slavaleleka (approved):

Alex-302 (approved):

Alex-302 (approved):

slavaleleka (approved):

PR #174493, opened 2024-03-09
add b-cdn.net to lockdown rules and lockdown (nsfw) everia.club

Yuki2718:

- [ ] Filters maintenance.

### Enter the issue address

As or writing, [NSFW] `everia.club` looks to be hijacked with malicious script in `data:text/javascript` which calls `https://dns.google/resolve?name=...` to redirect visitor to scam. I wanted to lockdown the site which anyway has PopAds, but then only to find such a common CDN of `b-cdn.net` is not included in the lockdown rules.

1. Your comment

### Terms


Yuki2718:
The injected code is mostly the same as `https://blog.sucuri.net/2023/08/from-google-dns-to-tech-support-scam-sites-unmasking-the-malware-trail.html` except for using data URI than atob and not using addEventListener. Whether the redirection to scam actually happens or not depends on various factors. I was redirected only on the first two visits. Of note the site was first reported by https://github.com/AdguardTeam/AdguardFilters/issues/89377

AdamWr (approved):

AdamWr:
It looks like that currently it uses `web-hosts.io` for redirection - `hxxps://web-hosts[.]io/?cnmnmtijvq3fk5e6e9s0` (this part after `/?` seems to depends on IP, but I'm not sure), so maybe `web-hosts.io` could also be blocked.

Yuki2718:
Sure, I'll block in EL.

Yuki2718:
The script lockdown here doesn't block the redirect as it is based on xhr so needing xhr lockdown too, is it worth a new rule? I also added https://github.com/uBlockOrigin/uAssets/commit/94f060d36d696458448036544777769f076516b0 to uBO, but according to `https://blog.sucuri.net/2023/08/from-google-dns-to-tech-support-scam-sites-unmasking-the-malware-trail.html` the malicious actor used `tracker-cloud.com` in past so probably this part will change.

AdamWr:
>The script lockdown here doesn't block the redirect as it is based on xhr so needing xhr lockdown too, is it worth a new rule?

There is already `$xmlhttprequest,third-party,denyallow=[...],domain=[...]` in Turkish filter, so if it block ads and doesn't cause any breakage, then I don't see any reason to not add similar rule to Base filter, I guess that it can be useful on other websites.

Alex-302 (approved):

PR #176701, opened 2024-04-10
fix adobe.com #176525

slavaleleka:
https://github.com/AdguardTeam/AdguardFilters/issues/176525

### DIFF
--- a/AnnoyancesFilter/Other/sections/self-promo.txt
+++ b/AnnoyancesFilter/Other/sections/self-promo.txt
@@ -137,6 +137,8 @@ www.kinopoisk.ru##div[class^="trailer-promo"]
 ! END: Yandex
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/176525
+adobe.com##.promobarModal
 ! https://github.com/AdguardTeam/AdguardFilters/issues/176017
 reports.adguard.com,reports.adguard.info##.flash-wrapper-wrap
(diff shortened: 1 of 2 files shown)


AnnoyancesFilter/Other/sections/self-promo.txt
```diff
@@ -137,6 +137,8 @@ www.kinopoisk.ru##div[class^="trailer-promo"]
 ! END: Yandex
 !
 !
+! https://github.com/AdguardTeam/AdguardFilters/issues/176525
+adobe.com##.promobarModal
```
Sergey-Lyapin, line 141: to popups maybe?

slavaleleka: even if it's actually a self-promo?

Sergey-Lyapin: as well as a popup

zloyden: I think it's ok to move it to popups.

slavaleleka: so 'popup' prevails over 'self-promo' in this case and the rule should be moved to 
`AnnoyancesFilter/Popups/sections/popups_specific.txt` ?

Alex-302: Yes.

zloyden (approved):

Sergey-Lyapin (approved):

PR #177755, opened 2024-04-24
💉 Enhance: #177742 Hide element at gyutte.jp

admitrevskiy:
Fix for issue https://github.com/AdguardTeam/AdguardFilters/issues/177442

### DIFF
--- a/JapaneseFilter/sections/specific.txt
+++ b/JapaneseFilter/sections/specific.txt
@@ -3035,6 +3035,7 @@ newsmomonga.com##.post div[align="center"] + br + hr
 nc-net.or.jp##.box-pr-image-01
 mori.nc-net.or.jp###ok_ad_side
 gyutte.jp###gt_sp_all_header
+gyutte.jp###gt_sp_all_footer
 jm1xtk.com##.web_adber-1
 allabout.co.jp,city.soka.saitama.jp,jm1xtk.com##div[class$="_ad"]
 my-best.com##.l-footer__fixed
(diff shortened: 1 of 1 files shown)


JapaneseFilter/sections/specific.txt
```diff
@@ -3034,6 +3034,7 @@ newsmomonga.com##.post div[align="center"] + br
 newsmomonga.com##.post div[align="center"] + br + hr
 nc-net.or.jp##.box-pr-image-01
 mori.nc-net.or.jp###ok_ad_side
+gyutte.jp###gt_sp_all_header
```
AdamWr, line 3037: It seems that there is another similar ad leftover at the bottom.

image

</details>

Could you please hide it too?
I guess that in this case it can be done by modifying current rule.
```adblock
gyutte.jp##div[id^="gt_sp_all_"]
```

AdamWr (changes requested):

Alex-302 (approved):

Alex-302:
@admitrevskiy Please also write here why used two rules instead of `gyutte.jp##div[id^="gt_sp_all_"]`

AdamWr (approved):

JapaneseFilter/sections/specific.txt
```diff
@@ -3034,6 +3034,9 @@ newsmomonga.com##.post div[align="center"] + br
 newsmomonga.com##.post div[align="center"] + br + hr
 nc-net.or.jp##.box-pr-image-01
 mori.nc-net.or.jp###ok_ad_side
+! We use 2 separate rules instead of gyutte.jp##div[id^="gt_sp_all_"] since the original reporter uses AdGuard for iOS and this rule does not apply in safari
```
Alex-302, line 3037: I mean in PR :harold:
```suggestion
```

Alex-302 (changes requested):

admitrevskiy:
> @admitrevskiy Please also write here why used two rules instead of `gyutte.jp##div[id^="gt_sp_all_"]`

! We use 2 separate rules instead of gyutte.jp##div[id^="gt_sp_all_"] since the original reporter uses AdGuard for iOS and this rule does not apply in safari

Alex-302 (approved):

PR #177839, opened 2024-04-24
💉 Enhance: #177828; Add a rule that prevents adding event listener fo…

admitrevskiy:
Fix for https://github.com/AdguardTeam/AdguardFilters/issues/177828

### DIFF
--- a/AnnoyancesFilter/Popups/sections/popups_specific.txt
+++ b/AnnoyancesFilter/Popups/sections/popups_specific.txt
@@ -15,7 +15,7 @@
 !
 ! SECTION: Popups - Regular rules
 !
-newslaundry.com#%#//scriptlet('prevent-addEventListener', 'scroll', 'function')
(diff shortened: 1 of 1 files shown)


AnnoyancesFilter/Popups/sections/popups_specific.txt
```diff
@@ -15,6 +15,7 @@
 !
 ! SECTION: Popups - Regular rules
 !
+newslaundry.com#%#//scriptlet('prevent-addEventListener', 'scroll', 'function')
```
AdamWr, line 18: If I'm not wrong, it will prevents all (or almost all) event listeners with `scroll` event, so maybe it could be a bit more specific:
```suggestion
newslaundry.com#%#//scriptlet('prevent-addEventListener', 'scroll', 'pageYOffset')
```

Alex-302 (approved):

PR #177842, opened 2024-04-24
Fix/177395

cannt39t:
#177395


BlazDT (changes requested):
To allowlist (free iOS and android content blockers do not support #%# rules):
`uhrzeit.org#@#.acceptCookies`

BlazDT (approved):

Alex-302 (approved):

PR #178389, opened 2024-05-02
fix: `pixiv.net` cookie notices and infinite reload due to rules conflict

JasonKhew96:

- [ ] Filters maintenance.

## What issue is being fixed?

- fix `pixiv.net` cookie notices rule, the cookie value should be `6`
- fix infinite reload if enabled with `EasyList - Cookie Notices` since this filter set cookie value to `5` and `EasyList - Cookie Notices` set cookie value to `6`

### Terms


Yuki2718 (approved):

Yuki2718:
@Alex-302 Hard breakage on a very popular site, it's not something to be left for days.

JasonKhew96:
context:
https://www.reddit.com/r/Pixiv/comments/1cia7rx/people_who_cant_load_pixiv_pixiv_doesnt_like/
https://www.reddit.com/r/Pixiv/comments/1chioso/i_finally_got_pixiv_to_load_but_why_is_it_using_a/
https://www.reddit.com/r/Pixiv/comments/1cg2bq5/pixiv_is_not_loading/

AdamWr (approved):

jackrjli:
This breakage will happen again if Pixiv changes the cookie value (which I guess they do when they update their privacy policy and need to notify everyone again) and the lists update to reflect that. I wonder whether it is necessary to set the cookie at all, since even without this rule the privacy policy update banner is hidden by a cosmetic filter:
```
pixiv.net##div > div:not([class]) > div[class^="sc-"]:has(> div > p > span > a[href^="https://policies.pixiv.net"])
```
https://github.com/AdguardTeam/AdguardFilters/blob/9bb7c0054eb436e51692b07db4f5baef7f82b5c0/AnnoyancesFilter/Cookies/sections/cookies_specific.txt#L6855

krystian3w:
If can by hidden instead `trusted-click-element` if button are stable.

AtomicLugia:
Pixiv is at it again, the filter doesn't work anymore, infinite reload is back

Yuki2718:
> Pixiv is at it again, the filter doesn't work anymore, infinite reload is back

I can't reproduce.

Catman-232:
Found my way here due to searching this issue, sorry for bumping. I don't remember how long it's been the case but this happens to me when going to gallery pages that have been deleted, which is a "Page not found" page, but on the original `/artworks/1234567etc` url. So for example this one that just got me thinking about this: https://www.pixiv.net/en/artworks/55438630

In uBO's logger I don't really know what to look for but I do see the `pixiv.net##+js(set-cookie-reload, privacy_policy_agreement, 6)` rule, followed by EasyList - Cookie Notices `pixiv.net##+js(set-cookie, privacy_policy_agreement, 6, , reload, 1)`

Actually even stranger is that I opened it in a private browsing window and it reloads over and over just the same, so maybe it's actually something about the web browser config or just something on Pixiv's end.

JasonKhew96:
> Found my way here due to searching this issue, sorry for bumping. I don't remember how long it's been the case but this happens to me when going to gallery pages that have been deleted, which is a "Page not found" page, but on the original `/artworks/1234567etc` url. So for example this one that just got me thinking about this: https://www.pixiv.net/en/artworks/55438630
> 
> In uBO's logger I don't really know what to look for but I do see the `pixiv.net##+js(set-cookie-reload, privacy_policy_agreement, 6)` rule, followed by EasyList - Cookie Notices `pixiv.net##+js(set-cookie, privacy_policy_agreement, 6, , reload, 1)`
> 
> Actually even stranger is that I opened it in a private browsing window and it reloads over and over just the same, so maybe it's actually something about the web browser config or just something on Pixiv's end.

@Catman-232 reproduced without any extension, with some early investigation, pixiv frontend try to redirect between `https://www.pixiv.net/artworks/55438630` and `https://www.pixiv.net/{language}/artworks/55438630` if you don't have Japanese language selected.

krystian3w:
> reproduced without any extension

So page logic is broken in that case.

PR #179359, opened 2024-05-16
Attempted to improve close-out of Regex dots

DandelionSprout:
# Creating the pull request

- [x] Filters maintenance.

## What issue is being fixed?

None at the moment, as reports.adguard.com has no options for any scenarios like this.

### Enter the issue address

None at the moment, as reports.adguard.com has no options for any scenarios like this.

1. Your comment

I had noticed that some entries had unclosed .'s in RegEx entries, in particular in cases where one would normally expect them to be enclosed unless there's something I'm currently unaware of, such as domain names and when leading into file extensions. So I attempted to fix that in this PR.

2. Screenshots

<details>

<summary>Screenshot 1:</summary>

image

</details>

### Terms


BlazDT (approved):
Makes sense to fix regex.

Alex-302:
It is not a [big] problem, but will make it harder to find domains in filters.

DandelionSprout:
I must say that that makes pretty much no sense, as it'd still be easy to find `briefmenow`, and also that all but 1 of the changed entries have `$domain` with the same domains as before.

PR #182166, opened 2024-06-24
Fix npo.nl

EchoDev:
This fixes the pre-roll adds shown on npo.nl

This can possibly work for other npo.nl related sites but are not tested.

### DIFF
--- a/DutchFilter/sections/specific.txt
+++ b/DutchFilter/sections/specific.txt
@@ -4,6 +4,7 @@
 ! Good: example.org###rek; ||example.com/ads/
 ! Bad: example.org#@##banner_ad (for instance, should be in allowlist.txt or in antiadblock.txt)
 !
+||azurewebsites.net^$domain=npo.nl
 ||biernet.nl/site/plugins/bja_new/redirect.php
 biernet.nl##.bja-cont
 biernet.nl###ban_blok


DutchFilter/sections/specific.txt
```diff
@@ -4,6 +4,7 @@
 ! Good: example.org###rek; ||example.com/ads/
 ! Bad: example.org#@##banner_ad (for instance, should be in allowlist.txt or in antiadblock.txt)
 !
+||azurewebsites.net^$domain=npo.nl
```
AdamWr, line 7: I guess it's about ad like reported here - https://github.com/AdguardTeam/AdguardFilters/issues/166205
If so, it's already fixed by a scriptlet and I cannot reproduce it, though it might be useful for platforms which do not support scriptlets.
I guess that `azurewebsites.net` might be used for other purposes than ad as well, so could this rule be more "ad related"?
For example:
```adblock
||azurewebsites.net/*adslot$domain=npo.nl
```

EchoDev: The scriptlet did not work for me on Firefox with uBlock Origin (all latest version).

`azurewebsites.net` is only used for tracking and serving ads from the network connections I saw, I noticed no breakage on the site itself

AdamWr: As far as I can see, `azurewebsites.net` can be used for other content as well.
Even though it doesn't break anything, I still prefer more specific rule(s), at least for now.

Alex-302:
https://github.com/uBlockOrigin/uAssets/issues/24275

PR #183319, opened 2024-07-10
DevUploads Update

BlazeFTL:
`devuploads.com/uo0kz55tm5h4`

The domains removed are dead
And rules are obsolete


Alex-302 (changes requested):
djxmaza.in is alive

BlazeFTL:
Yes But The Filters Are Not Required Or Obsolete

Alex-302 (approved):

PR #183363, opened 2024-07-10
Fix #183353 www.vox.com

rakleed:
Fix #183353 www.vox.com


AnnoyancesFilter/Popups/sections/subscriptions_specific.txt
```diff
@@ -54,6 +54,8 @@ armidaleexpress.com.au,bendigoadvertiser.com.au,canberratimes.com.au,centralwest
 !
 ! SECTION: Subscriptions - Regular rules
 !
+vox.com##li:has(> a[href="/newsletters"])
```
BlazDT, line 57: Not a popup. User has to press on Menu to see it.
Imho move to Annoyances -> Other -> Annoyances.

Else pull request looks fine to me. Unless colleagues disagree about rule move.

Alex-302: image
It should not be blocked at all.

Alex-302:
```suggestion
```

Alex-302 (approved):

PR #189978, opened 2024-09-30
Address https://github.com/AdguardTeam/AdguardFilters/issues/189554#i…

Yuki2718:

- [ ] Filters maintenance.

## What issue is being fixed?

https://github.com/AdguardTeam/AdguardFilters/issues/189554
https://github.com/AdguardTeam/AdguardFilters/issues/189663
https://github.com/AdguardTeam/AdguardFilters/issues/189883
etc.

1. Your comment

The current rules are nonsense on iOS where `$redirect` is not supported. Though simply blocking `html-load.com/loader.min.js` is suffice, doing so is problematic for free user without advanced protection. Hence `prevent-element-src-loading`.

### Terms


Yuki2718:
I still worry `remove-node-text` may be too slow to execute on iOS, but I don't have any means to test.

Alex-302 (approved):

Yuki2718:
> I still worry `remove-node-text` may be too slow to execute on iOS, but I don't have any means to test.

Best to watch detection reports - can happen/increase if the above scenario is the case.

Issue #192475, opened 2024-11-07
AdGuard Base Filter breaks user rules at www.ebay.com


kriswilk:
### Prerequisites

- [X] This site DOES NOT contain sexually explicit material, otherwise use [NSFW-specific form](./new?template=bug_report_NSFW.yml);
- [X] Filters were updated before reproducing an issue;
- [X] AdGuard product version is up-to-date;
- [X] Browser version is up-to-date;
- [X] If the site or app is broken, disabling AdGuard protection resolves an issue.

### What product do you use?

AdGuard DNS, AdGuard for Windows

### AdGuard version

7.19.0.4853

### What type of problem have you encountered?

AdGuard gets detected on a website

### Which browser(s) do you use?

Chrome

### Which device do you use?

Desktop

### Where is the problem encountered?

https://www.ebay.com/itm/186333967862 (or any other live listing)

### Ad Blocking

AdGuard Base filter

### Privacy

AdGuard Tracking Protection filter

### Social Widgets

AdGuard Social Media filter

### DNS filters

AdGuard DNS filter

### Add your comment and screenshots

_Issue #192269 was closed without resolving the actual problem. I don't know if maintainers will see my subsequent comments there, so I created this much more detailed issue report._

### SUMMARY

Particular user rules for www.ebay.com break only when the "AdGuard Base Filter" is enabled.

Follow the steps below to reproduce the issue. Conducting the same test with the same user rule in uBO works fine. This is specific to AdGuard Base Filter.

### STEPS TO REPRODUCE:

1. Make sure there are NO user rules
2. Make sure "AdGuard Base FIlter" is ENABLED
3. DISABLE AdGuard
4. Visit https://www.ebay.com/itm/186333967862 ... promoted content exists within a `<div>` with class `x-rx-slot`

<details><summary>Screenshot</summary>

![Screenshot 2024-11-07 092939](https://github.com/user-attachments/assets/3e816171-fa2f-4d03-b1ce-7d0228dbc998)

</details>

5. ENABLE AdGuard
6. Reload the test page ... promoted content still exists within a `<div>` with class `x-rx-slot`

<details><summary>Screenshot</summary>

![Screenshot 2024-11-07 093042](https://github.com/user-attachments/assets/8f3122af-82c6-428e-80b3-2e8217a2d3cc)

</details>

7. Now create a user rule `ebay.*##.x-rx-slot`
[...]

PR #192487, opened 2024-11-07
Remove filter for "useful" content at ebay.*

kriswilk:
Issue #192475 was addressed by @BlazDT in https://github.com/AdguardTeam/AdguardFilters/commit/167a7b0f6659ee9500d9fa53c23963d09bf1e472. That commit:

- adds anti-adblock entries (thank you!!)

but also

- adds a filter for the unwanted content that I was targeting
- adds a "usefulads" exclusion for the same content

Since many users probably enjoy using the "similar items" and "you may also like..." sections, this should NOT be the default behaviour.

For that reason, this PR removes the filter rule and its "usefulads" counterpart. The anti-adblock component was all that was required to solve the original issue.

### DIFF
--- a/BaseFilter/sections/specific.txt
+++ b/BaseFilter/sections/specific.txt
@@ -388,7 +388,6 @@ $xmlhttprequest,third-party,denyallow=fastlycdn.com|doodcdn.co|intensedebate.com
 !
 ! NOTE: Specific rules
 !
-ebay.*##.x-rx-slot
 sunderlandecho.com###first-sticky-group
 sunderlandecho.com##.taboola-bottomfeed
 sunderlandecho.com##div[id^="leaderboard"]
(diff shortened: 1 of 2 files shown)


BlazDT:
They are sponsored.
So if a user wants to see them just enable useful ads filter

kriswilk:
> They are sponsored. So if a user wants to see them just enable useful ads filter

OK. In that case, may I suggest a more complete set of filters that strip out obvious sponsored content from eBay? This is what I use:

```
! ebay listings - promoted content carousels
ebay.*##.x-rx-slot
ebay.*##.x-rx-slot-btf
! ebay search results - picked for you, related searches
ebay.*##.srp-river-answer
ebay.*##.srp-main-below-river
```

Can you add these, or shall I make a new PR?

BlazDT:
`.x-rx-slot-btf` - `hides You may also like` - not a sponsored element

Modified first search results rule (else arrows to continue to next page are gone):
ebay.*##.srp-river-answer[class*="CAROUSEL"]

Last rule is not sponsored.

kriswilk:
I appreciate the quick reply, the correction of my rule (I never noticed the missing arrows!), and for being cautious about elements not specifically labelled as "sponsored". That said, the others are just as annoying/spammy...and they are all fed to us by eBay. 🤷‍♂️

In any case, I'll continue to filter out any remaining junk myself. Thank you for the anti-adblock fix!

BlazDT:
@kriswilk
Had to modify one rule again (was hiding filtering buttons on mobile devices).

https://github.com/AdguardTeam/AdguardFilters/commit/53e4cf5aa41d293069061e76247539654c0acabf

PR #196052, opened 2025-01-09
Problematic rules AdguardFilters/issues/196022

Alex-302:
https://github.com/AdguardTeam/AdguardFilters/issues/196022

### DIFF
--- a/.aglintrc.yaml
+++ b/.aglintrc.yaml
@@ -9,6 +9,8 @@ rules:
     - error
     - regexp-patterns:
         # Do not forget to escape special regex characters
+        - ^\|\|b\.karte\.io
+        - ^\|\|static\.karte\.io
         # https://github.com/AdguardTeam/AdguardFilters/commit/a64fb16a9c0ba926f980ca5091ca6d39eeb6ef40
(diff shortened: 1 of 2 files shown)


.aglintrc.yaml
```diff
@@ -9,6 +9,9 @@ rules:
     - error
     - regexp-patterns:
         # Do not forget to escape special regex characters
+        # https://github.com/AdguardTeam/AdguardFilters/issues/191068#issuecomment-2480464336
+        - ^\|\|b\.karte\.io
+        - ^\|\|static\.karte\.io
```
zloyden, line 14: I guess if the problem is widely known, it can be removed, so I agree with that.

Yuki2718 (approved):

zloyden (approved):

Yuki2718:
Theoretically, `$removeparam` or such a filter may be added in future and should not be blacklisted?

Alex-302:
> Theoretically, `$removeparam` or such a filter may be added in future and should not be blacklisted?

Ok, we will check removeparam

Yuki2718:
Probably domain-specific rule should be exempted. Sorry, didn't notice in the review.

PR #199767, opened 2025-03-01
Upd #198988 decrypt.day

rakleed:
It's one of the triggers of the anti adblock script.

You can manually disable these rules and add uBlock filters (https://raw.githubusercontent.com/uBlockOrigin/uAssetsCDN/refs/heads/main/filters/filters.txt), then the anti adblock script won't be triggered.


Alex-302:
I can't reproduce.

PR #201559, opened 2025-03-24
Fix ufret.jp ad-shield

ntnguyen1234:
URL: https://www.ufret.jp/

Also removed `ndtvprofit.com` from ad-shield section since they no longer uses it I think. Please check again to be sure.


Yuki2718:
I don't see AS ads on the site though.
> Also removed `ndtvprofit.com` from ad-shield section since they no longer uses it I think. Please check again to be sure.

I have seen some sites "paused" AS ads and then later do inject again.

ntnguyen1234:
Maybe due to generic cosmetic rules? I can see the ad-shield elements (start with `img_`) at the bottom of a leftover:

image

Alex-302 (approved):

PR #202132, opened 2025-04-02
Upd dizipal sites

ntnguyen1234:
From: https://www.reddit.com/r/uBlockOrigin/comments/1jp1jnj/dizipal951com_not_working/

From what I test, looks like we can use these rules instead of trusted scriptlets. Can anyone confirm?


AdamWr:
It works fine on `dizipal951.com` but unfortunately it breaks video player on `https://dizipal2.com/gibi-1-sezon-12-bolum-izle-full-4` (at least on my end).

ntnguyen1234:
Looks like the ones with small number are the fake dizipal (`! dizipal*.com - fake dizipal`). How about `[$domain=/dizipal\d{3,}.com/]` ?

AdamWr:
I guess we can try, though current rule works fine for both cases.

ntnguyen1234:
Hmm... Then I think we can keep that. I'll update the domain numbers since they are over 950 now.

ntnguyen1234:
Oh sorry, I copy wrong rules. I'll update them.

AdamWr:
Could you please add `.com` to all new rules?

ntnguyen1234:
Thanks. Updated.

AdamWr (approved):

AdamWr:
Thank you.

zloyden (approved):

ntnguyen1234:
Thanks. I forgot about twitter link. If I'm not wrong, regex domains are not supported in MV3 and AG content blocker (on samsung, yandex... browsers).

zloyden (approved):

AdamWr (approved):

Alex-302 (approved):

PR #203436, opened 2025-04-18
Upd #203429 grapee.jp

ntnguyen1234:
Update https://github.com/AdguardTeam/AdguardFilters/issues/203429

Ad-Shield.


Yuki2718:
Not sure if we have to add the domain to TPF when already fully blocked in Base.

Yuki2718:
It's a very high-traffic site. I'll fix directly.

PR #209406, opened 2025-07-15
Maintenance: Remove non-removeparam rules from URL tracking param filter

rozbb:
The URL tracking parameters filter has rules in it that do not remove any tracking parameters, i.e., they don't have a `removeparam` modifier. I noticed this while writing a parser for the [compiled list](https://github.com/AdguardTeam/FiltersRegistry/blob/master/filters/filter_17_TrackParam/filter.txt).

This PR removes these unnecessary rules. I'm pretty sure I got all of them. I just ran `:v/\(^!\|removeparam\)/d` in Vim, and cleaned it up after.

This doesn't have a corresponding issue.

### Terms

- [X] By submitting this issue, I agree that pull request does not contain private info and all conditions are met

### DIFF
--- a/TrackParamFilter/sections/general_url.txt
+++ b/TrackParamFilter/sections/general_url.txt
@@ -104,8 +104,6 @@ $removeparam=sms_uph
 $removeparam=ttclid
 ! spot.im
 $removeparam=spot_im_redirect_source
-! WordPress Visitor Statistics plugin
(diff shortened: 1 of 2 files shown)


Alex-302 (approved):

Alex-302:
Thanks, will add to the correct section.

PR #211904, opened 2025-08-22
Fix #211881 www.rgshows.me

xglossyntx:
# Creating the pull request

- [ ] Filters maintenance.

### Enter the issue address

Fix #211881
Fix #211882

### Terms


AnnoyancesFilter/Other/sections/tweaks.txt
```diff
@@ -1668,6 +1668,7 @@ t24.com.tr#%#//scriptlet('set-constant', 'Object.prototype.autoplay', '0')
 /wp-content/plugins/disable-right-click/*
 /wp-content/plugins/right-click-disable-orignal/*
 !
+||cdn.jsdelivr.net/npm/disable-devtool
```
Sergey-Lyapin: please make this rule domain-specific

BaseFilter/sections/specific.txt
```diff
@@ -723,8 +723,8 @@ $third-party,script,domain=klto9.com|f2movies.to|hacamchicac.com|yourporngod.com
 !
 ! NOTE: Specific rules
 !
-9to5toys.com###openweb-pitc
```
Sergey-Lyapin, line 726: why did you delete these rules?

Sergey-Lyapin (changes requested):

BaseFilter/sections/specific.txt
xglossyntx: I'm not sure why it was removed, but updating it again seems to have fixed the issue.

AnnoyancesFilter/Other/sections/tweaks.txt
xglossyntx: Does it break something? Because according to AdGuard's filter policy, it should be blocked on all sites, seeing as its only goal is to disable browser features such as DevTools and the context menu.

https://adguard.com/kb/general/ad-filtering/filter-policy/#adguard-other-annoyances-filter
`Unblock actions on the page, such as opening the context menu, selecting and copying text, if they are blocked`

AdamWr: We had similar rule, but it has been removed because some websites were broken - https://github.com/AdguardTeam/AdguardFilters/commit/2be2f73f9688785c44b2ead549b05760c247d459#diff-15e8c7212f7e9ee18ff5e97a5cdb569d7df243071f1ff9a62cf3f456f67c414dL1812
So if it's really required, then domain specific rule is preferable.

xglossyntx:
This should be ready to merge.

Sergey-Lyapin (approved):

PR #212066, opened 2025-08-24
Fix Web Share API

piquark6046:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Share_API


Alex-302:
Isn't it a legal feature?
It seems is can't be used without user permission 
image

piquark6046:
Please take a look with macOS.

<p>

Screenshot 2025-08-27 at 10 29 25 PM


</p>
</details>

Alex-302:
And? We block social sharings. But this feature is common (adding to notes etc)

PR #216348, opened 2025-10-27
Removed

xglossyntx:
Removed


Alex-302:
Hm, why closed?

xglossyntx:
The rule did not work as intended

PR #217447, opened 2025-11-10
Fix stale issue message in GH Actions

rakleed:
See https://github.com/AdguardTeam/AdguardFilters/issues/217105#issuecomment-3510327488


AdamWr:
Fixed in https://github.com/AdguardTeam/AdguardFilters/commit/6509c162c6047f861ee8e351740e778311410a53

PR #219859, opened 2025-12-11
update Nodejs version to 24

piquark6046:
Apply the latest LTS version of Nodejs.


scripthunter7 (approved):

Alex-302 (approved):

piquark6046:
@scripthunter7 https://github.com/8398a7/action-slack is no longer be maintained.
I think that we should find an alternative.

PR #222977, opened 2026-01-18
fix #222976

Khrin:
https://github.com/AdguardTeam/AdguardFilters/issues/222976


Khrin:
(I'm sorry, I couldn't think of anything better)

SocialFilter/sections/specific.txt
```diff
@@ -87,6 +87,8 @@ cogconnected.com,echofm.online,antirez.com,comicbook.com,alexp.pl,2ality.com,sma
 !
 ! NOTE: Specific rules
 !
+iranintl.com##.RelatedArticleCard-module-scss-module__40K_PG__card__shareButton
```
xglossyntx: This one does not look like a social share button. It just copies the URL and article headline to your clipboard.

SocialFilter/sections/specific.txt
```diff
@@ -87,6 +87,8 @@ cogconnected.com,echofm.online,antirez.com,comicbook.com,alexp.pl,2ality.com,sma
...
+iranintl.com##.URLShareDialog-module-scss-module__YDVwKa__shareButton
```
Alex-302: We usually add more wide selectors, because on some sites parts like `__40K_PG__` / `__YDVwKa__` are changed from time to time.

```suggestion
iranintl.com##button[aria-label="common.share"]
```

or 

```
button[class$="__shareButton "]
```

SocialFilter/sections/specific.txt
Alex-302: > It just copies the URL and article headline to your clipboard.

not a problem if it is a part of a widget. Anyone can copy URL. Same with a Print buttons.

PR #224207, opened 2026-02-05
fix #224203 parents.com

ghajini:
# Creating the pull request

- [ ] This is not an ad/bug report;
- [ ] My code follows the [guidelines](https://github.com/AdguardTeam/AdguardFilters/blob/master/CONTRIBUTING.md) and [syntax](https://kb.adguard.com/general/how-to-create-your-own-ad-filters) of this project;
- [ ] I have performed a self-review of my own changes;
- [ ] My changes do not break web sites, apps and files structure.

- [ ] Filters maintenance.

### Enter the issue address

 Fix <https://github.com/AdguardTeam/AdguardFilters/issues/224203>

### Terms


BaseFilter/sections/specific.txt
```diff
@@ -188,8 +188,8 @@ hookemheadlines.com,12thmanrising.com,aroundthefoghorn.com,arrowheadaddict.com,b
...
...bakedharvest.com,wellplated.com,platingsandpairings.com,foxeslovelemons.com,crunchycreamysweet.com,backforseconds.com,cookingclassy.com,brighteyedbaker.com,hoosierhomemade.com,entertainingwithbeth.com,tidefans.com,butterbeready.com,nourishedbynic.com,juliasalbum.com,fromvalerieskitchen.com,iamhomesteader.com,kdnuggets.com,androidheadlines.com,insider-gaming.com,juliescafebakery.com,ohsheglows.com,averiecooks.com,nomnompaleo.com,healthyfitnessmeals.com,redcurrantbakery.com,the-girl-who-ate-everything.com,stylegirlfriend.com,acedarspoon.com,addapinch.com,amandascookin.com,aprettylifeinthesuburbs.com,bakedbree.com,bakedbyrachel.com,barbarabakes.com,feastingathome.com,feedmephoebe.com,framedcooks.com,healthylittlefoodies.com,hungrypaprikas.com,ibreatheimhungry.com,littlesunnykitchen.com,melskitchencafe.com,mexicanfoodjournal.com,mountainmamacooks.com,mycolombianrecipes.com,neighborfoodblog.com,seededatthetable.com,shugarysweets.com,tasteandtellblog.com,thebakermama.com,theendlessmeal.com,thefoodieaffair.com,themediterraneandish.com,thewanderlustkitchen.com,whatjewwannaeat.com,gearpatrol.com,studyfinds.org,unefemme.net,hummingbirdhigh.com#%#//scriptlet('prevent-xhr', 'ads.adthrive.com')
```
Alex-302: are these two rules required? I don't see `__adblocker` cookie in scripts.
Also I don't see unblocked ads. Maybe the rest are useful, but not sure about this pair.

ghajini: ok will remove as per review

BaseFilter/sections/specific.txt
```diff
@@ -107,16 +107,16 @@ weatherzone.com.au,lifehacker.jp#%#//scriptlet('trusted-replace-node-text', 'scr
...
+parents.com,merinfo.se,crosswordsolver.com,natalie.mu,petbook.de,egitim.net.tr,9gag.com,manchesterworld.uk,watchdocumentaries.com,canuckaudiomart.com,windsorstar.com,pons.com,bz-berlin.de,myhomebook.de,fitbook.de,techbook.de,travelbook.de,twopeasandtheirpod.com,dict.cc,fansided.com,nationalpost.com,zeilen.nl,wielerrevue.nl,ukaudiomart.com,woxikon.de,wort-suchen.de,happyinshape.com,thetruthaboutcars.com,foodandwine.com,tips-and-tricks.co,lejdd.fr,edinburghnews.scotsman.com,sussexexpress.co.uk,lerobert.com,ldoceonline.com,yorkshirepost.co.uk,seriouseats.com,mamastar.jp,collinsdictionary.com,tipsandtricksjapan.com,homesicktexan.com,oxfordlearnersdictionaries.com,dictionary.cambridge.org,bab.la,liveandletsfly.com,kuruma-news.jp,infinityfree.com,autobild.de,toyokeizai.net,usaudiomart.com,songfacts.com,newatlas.com,fossbytes.com,perezhilton.com,exceljet.net,sportico.com,wwd.com,indiewire.com,helpdeskgeek.com,akinator.com,dnevnik.hr,rollingstone.com,billboard.com,variety.com,hollywoodreporter.com,nouvelobs.com,deadline.com,computerbild.de,divinedaolibrary.com,thelayoff.com,welt.de,scribd.com,boardingarea.com,doviz.com,oricon.co.jp$$script[src="html-load.com/"]
```
AdamWr: Do you see it as inline script?
On my end it's loaded from this script:
```
https://www.parents.com/static/3.195.0/cache/eNqlVVty2zAMvFBRTY7Q6W__egKIhCg4FKkhIDvO6QPRduJXOOr0Q0Nwyd0BQEDoRFHZdRMmjdTtpJvMgD0mjhFBXMkx_uhub7k8zTlRUuk4KZWEsfM0UFn51SjkP0mGOSnDqsWzXsNBJ8iz8sTvVK4PLhqAJn00UL5cyH6JJNc-aHkDHAaOjEqgGAKVezRyeoVCh8J6ctP4ExVHrXuNuOntFLd5Xjmyaj6A_y0AlglcosIcl8DpIQvG-eXh78gUfYcepFrPrkV8PwL6hkcH9oFsHYnDqL9HTIH-sCgly9gzsKEVYu6tKEZCX8nrcn89cl89w5eXI3jGmENDMeGeg53l1C1qj6VHMOjabpCHXCbphLC4EdbNtd3gTSSCgaDHtCbhdtvgnaI5L_Ult0U2LDFamxClGtztdsPTndoV5pJDMV_N0fIM26CUsvLArrp1CX_3FG51JvqRhPf1v3Kxt6atlTE6SCS1XjnXzSPyyX4Sm1qbgWaQmciNd9tWOHv2lGF3mCMe13ycje8KO8rPPgjpd-di2OuxZ5Vt1aHZnTlwa26nL8leoAhGuBJ6Bm6XtA8mLK-WkC_zn-it8TJZv9VRwUPjmmhZnC7rxHDZJlLSTrG3O1-h1f2Gwteco_J8WRuMHoVOP-0Zk1_loY_ZtVrdsZ5aR_Jis6dWASf7PxBcjqRVuXvGwZgIdXpVRQuX05IXOc_qz6OWG_ekR-QDmGAeFw.min.js
```
If it's not inline script then these rules are not needed and it could be added only to this rule:
```adblock
#%#//scriptlet('trusted-suppress-native-method', 'Element.prototype.setAttribute', ' |"error-report.com"')
```
so it would be available for all platforms which support scriptlets.

By the way, we usually don't need to add it in both sections, one should be enough.

ghajini: It is rare to reproduce(error-report page)
used firefox android nightly & stable.
But ok trusted-suppress-native-method will work anyway even if its inline, i'll.update the pull accordingly

AdamWr (approved):

BaseFilter/sections/specific.txt
AdamWr: Approved, but I meant that we could just add one rule for this case. Other rules are rather not required until they change how this ad script is loaded.

zloyden (approved):

Alex-302 (approved):

PR #225452, opened 2026-02-23
Update tracking_servers.txt to allow `app.posthog.com`

rafaeelaudibert:
# Creating the pull request

Added wildcard entry for i.posthog.com and removed app.posthog.com.

Posthog never collects analytics via `app.posthog.com`, that's merely the entrypoint to the PostHog app used to route people between `us.posthog.com` and `eu.posthog.com`. All analytics is done via `*.i.posthog.com` instead.

The `$thirdparty` rule was supposed to help with this, but this is being added to DNS entries in https://github.com/AdguardTeam/AdGuardSDNSFilter and they don't support `$thirdparty`.

Another possible change is, instead of doing this, to simply add `app.posthog.com` as an exception in https://github.com/AdguardTeam/AdGuardSDNSFilter but that felt more wrong since it might raise some eyebrows.

See https://github.com/AdguardTeam/AdGuardSDNSFilter/issues/1660

- [ ] Filters maintenance.

### Enter the issue address

See https://github.com/AdguardTeam/AdGuardSDNSFilter/issues/1660

### Terms

### DIFF
--- a/SpywareFilter/sections/tracking_servers.txt
+++ b/SpywareFilter/sections/tracking_servers.txt
@@ -263,6 +263,7 @@
 ||cloudfilt.com^$third-party
 ||highway.cablecar.sph.com.sg^
 ||i.posthog.com^
+||*.i.posthog.com^
 ||app.optibase.io^$third-party
 ||abtshield.com^$third-party
 ||track.fusionmedia.io^
@@ -990,7 +991,6 @@
 ||px*.mediahills.ru^
 ||sdk.conscent.in^$third-party
 ||tpstelemetry.tencent.com^
-||app.posthog.com^$third-party
 ||crossees.com^
 ||humanz.com^$third-party
 ||gateway.foresee.com^
--- a/SpywareFilter/sections/tracking_servers.txt
+++ b/SpywareFilter/sections/tracking_servers.txt
@@ -263,7 +263,7 @@
 ||cloudfilt.com^$third-party
 ||highway.cablecar.sph.com.sg^
 ||i.posthog.com^
-||*.i.posthog.com^
+.i.posthog.com^
 ||app.optibase.io^$third-party
 ||abtshield.com^$third-party
 ||track.fusionmedia.io^


SpywareFilter/sections/tracking_servers.txt
```diff
@@ -263,6 +263,7 @@
 ||cloudfilt.com^$third-party
 ||highway.cablecar.sph.com.sg^
 ||i.posthog.com^
+||*.i.posthog.com^
```
Alex-302: `||*` is redundant in this case.

```suggestion
.i.posthog.com^
```

SpywareFilter/sections/tracking_servers.txt
```diff
@@ -990,7 +991,6 @@
 ||px*.mediahills.ru^
 ||sdk.conscent.in^$third-party
 ||tpstelemetry.tencent.com^
-||app.posthog.com^$third-party
```
Alex-302, line 993: I see on many sites the tracking script. That's why it is blocked by DSN.

https://app.posthog.com/static/array.js

It seems collects some ID's
```
    var rs = "https?://(.*)",
        ss = ["gclid", "gclsrc", "dclid", "gbraid", "wbraid", "fbclid", "msclkid", "twclid", "li_fat_id", "igshid", "ttclid", "rdt_cid", "epik", "qclid", "sccid", "irclid", "_kx"],
        ns = Fi(["utm_source", "utm_medium", "utm_campaign", "utm_content", "utm_term", "gad_source", "mc_cid"], ss),
        os = "<masked>",
        as = ["li_fat_id"];
```

Example
https://urlscan.io/result/019c8732-d281-745f-944c-21e3e556e599/#transactions

rafaeelaudibert: Huh, but they are using the wrong domain, since it should always be `us.i.posthog.com`or `eu.i.posthog.com`.

This would be totally fine to block if `$third-party` was doing its job - and it does when blocking in the browser - but it's just incorrect when blocking at the DNS level.

Maybe another option here is to block `/static/array.js`?

I'm happy to find a way to block it if we can guarantee `app.posthog.com` is unblocked and people will be able to use the app.

rafaeelaudibert: Ping @Alex-302

Alex-302: I'm checking.

Alex-302: https://github.com/AdguardTeam/AdguardFilters/commit/a93eb9df50ca7cff09cb54ef2ea9568e93cf7657

rafaeelaudibert: `array.js` might also be served by us.i.posthog.com or eu.i.posthog.com rather than just `app.posthog.com`.

Anyway, that's good! Thank you! Should we close this one?

Alex-302: yep. app.posthog.com is not blocked by our DNS now.

rafaeelaudibert: Thank you!

PR #226759, opened 2026-03-11
Fix #226645 interstyl-sklep.pl

mu-hun:
# Creating the pull request

- [ ] Filters maintenance.

### Enter the issue address

Fix #226645 interstyl-sklep.pl

`ssl.ceneo.pl/shops/sw.js` script loads the third-party review widget (floating promo/review widget embedded via iframe).

<details>
<summary>Screenshots</summary>

| Small widget | Extended widget |
| :--: | :--: |
| Screenshot 2026-03-11 at 12 40 00 PM | Screenshot 2026-03-11 at 12 39 50 PM |

</details>

So, I added the rule with `$third-party,domain` modifiers:

```adblock
||ssl.ceneo.pl/shops/sw.js^$third-party,domain=interstyl-sklep.pl
```

### Terms

### DIFF
--- a/AnnoyancesFilter/Cookies/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/Cookies/sections/cookies_specific.txt
@@ -397,6 +397,8 @@ parkhoteltokyo.com,novotel-danang-premier.com,swissotelnankaiosaka.com,pullmanto
 hotelnikkochangshu.com,shibaparkhotel.com###macaron_cookie_box
 txt-official.jp#%#//scriptlet('trusted-set-cookie', 'hbjOfficialSiteAcceptCookie', 'HBJ.1.0.0.0')
 ichiyoshi.co.jp##.popup_cookie_policy
+interstyl-sklep.pl##.consents
+interstyl-sklep.pl#%#//scriptlet('remove-attr', 'inert')
 googlegaido.quest,androidjapan.one###cookiesbox
 paulpadel.com##.con-kit-component-cookie-banner__backdrop
 profinance.ru##.askbuttontextbox
--- a/AnnoyancesFilter/Cookies/sections/cookies_specific.txt
+++ b/AnnoyancesFilter/Cookies/sections/cookies_specific.txt
@@ -397,8 +397,6 @@ parkhoteltokyo.com,novotel-danang-premier.com,swissotelnankaiosaka.com,pullmanto
 hotelnikkochangshu.com,shibaparkhotel.com###macaron_cookie_box
 txt-official.jp#%#//scriptlet('trusted-set-cookie', 'hbjOfficialSiteAcceptCookie', 'HBJ.1.0.0.0')
 ichiyoshi.co.jp##.popup_cookie_policy
-interstyl-sklep.pl##.consents
-interstyl-sklep.pl#%#//scriptlet('remove-attr', 'inert')
 googlegaido.quest,androidjapan.one###cookiesbox
 paulpadel.com##.con-kit-component-cookie-banner__backdrop
 profinance.ru##.askbuttontextbox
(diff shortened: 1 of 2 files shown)


AnnoyancesFilter/Widgets/sections/widgets.txt
```diff
@@ -1527,6 +1527,7 @@ blog.livedoor.jp##div[id^="shitarss_title_"]
 ||verbox.ru^$third-party
 ||uberlaw.ru^$third-party
 ||tidiochat.com^$third-party
+||ssl.ceneo.pl/shops/sw.js^$third-party,domain=interstyl-sklep.pl
```
AdamWr: It's a section for "generic" rules and this rule is domain specific, so could you please move it to "specific" section?
https://github.com/AdguardTeam/AdguardFilters/blob/7d23948a18f61e223febfb7d711eadcaea3cd528/AnnoyancesFilter/Widgets/sections/widgets.txt#L1551-L1553

AnnoyancesFilter/Cookies/sections/cookies_specific.txt
```diff
@@ -395,6 +389,8 @@ parkhoteltokyo.com,novotel-danang-premier.com,swissotelnankaiosaka.com,pullmanto
 hotelnikkochangshu.com,shibaparkhotel.com###macaron_cookie_box
 txt-official.jp#%#//scriptlet('trusted-set-cookie', 'hbjOfficialSiteAcceptCookie', 'HBJ.1.0.0.0')
 ichiyoshi.co.jp##.popup_cookie_policy
+interstyl-sklep.pl##.consents
+interstyl-sklep.pl#%#//scriptlet('remove-attr', 'inert')
```
AdamWr: One thing, scriptlets are not supported by all platforms (the same about CSS rules - `#$#`), so in a such case `#$#` rule should be used:
```adblock
interstyl-sklep.pl#$#.consents { display: none !important; }
```
This way website will be not broken when scriptlet is not applied.

But I think that in this case you can just add it to the current rule:
```adblock
#%#//scriptlet('set-local-storage-item', 'show_consent_modal', '1')
```
https://github.com/AdguardTeam/AdguardFilters/blob/ca3ec514a7b81fe10ee891e3f774d1977ca0dcdb/AnnoyancesFilter/Cookies/sections/cookies_specific.txt#L3094

AdamWr (changes requested):

AnnoyancesFilter/Cookies/sections/cookies_specific.txt
mu-hun: > One thing, scriptlets are not supported by all platforms (the same about CSS rules - #$#), so in a such case #$# rule should be used:

```adblock
interstyl-sklep.pl##.consents
```

The cosmatic rule syntax – `##.classname` support all platforms. aren't they?

zloyden: https://github.com/AdguardTeam/AdguardFilters/pull/226759#discussion_r2916497189 this rule breaks the website. However, it's fine with what `@Adam` suggested. Please also note that you shouldn't add any rules for the website, because it has low traffic, but we have the common one, so it's the exception.

mu-hun: @zloyden Ok, I got it.

AdamWr: Yes, it's supported by all platforms.
In this case, the problem is that `inert` attribute causes that elements on the website are not interactive.
So if `interstyl-sklep.pl##.consents` is added, then website will be broken on platforms that do not support scriptlets (iOS app free version and AdGuard Content Blocker).
In general, for such cases, we can use `#$#` rule instead, because it's also not supported by these platforms.

AdamWr (approved):

AnnoyancesFilter/Cookies/sections/cookies_specific.txt
```diff
@@ -3138,7 +3131,7 @@ tiesas.lv,portal.ccscollect.co.uk,jekabpils.lv,cvk.lv,seplp.lv,tieslietupadome.l
...
+aventurasport.pl,techberg.pl,bergson.pl,4air.pl,montres.pl,forkids.pl,bikeway.pl,wertykal.com,sport2002.pl,mikrobot.pl,sklep.amw.com.pl,mazur-sc.pl,software-klucze.pl,vobmat.pl,reolink-sklep.pl,pleciona.pl,abfoto.pl,slc.com.pl,znakidrogowe24.pl,traktorki.com,sklepdrogowy.pl,sklep.slawex24.pl,skarpetkowyszop.pl,shop.tagatic.com,nadrogi.pl,ms70.pl,filatelistyczny.pl,drogbit.pl,balneokosmetyki.pl,dlabrodacza.pl,planszostrefa.pl,arenalazienek.pl,ogrodolandia.pl,traumakit.pl,aedmax24.pl,mikavi.sklep.pl,wideorejestratory24.pl,wideorejestratory.com,zegarkinareke.pl,huzaro.pl,militarymarket.pl,r-kom.pl,csk-med.pl,filtrdodomu.pl,biosklep.com.pl,fhgerman.pl,deluxury.pl,sklepsoltech.pl,hanzo.com.pl,kozak.pl,trigar.pl,traffx.pl,luke.pl,wkdzik.pl,modnakiecka.pl,nkk.pl,pancernik.eu,gps.lodz.pl,bikepark24.pl,teesa.pl,sklepzycia.pl,interstyl-sklep.pl#%#//scriptlet('set-local-storage-item', 'show_consent_modal', '1')
```
zloyden: Please add it to the beginning of the rule. Also revert the original rule.

zloyden (approved):

mu-hun:
Rebased to avoid merge commits since the changes overlap with the master branch.

Alex-302 (approved):

PR #228232, opened 2026-04-01
Fix #228225 smplu.link

xglossyntx:
# Creating the pull request

- [ ] Filters maintenance.

### Enter the issue address

Fix #228225

### Terms

### DIFF
--- a/BaseFilter/sections/specific.txt
+++ b/BaseFilter/sections/specific.txt
@@ -1083,9 +1083,11 @@ thestreet.com##div[class^="styles_headerAdSlot__"]
 primesrc.me#%#//scriptlet('remove-attr', 'href', 'a.splash')
 primevid.click##a[href^="https://fairscrew.com/"]
 2short.club##.mdlAd
+mtpro.xyz##a[href]:contains(/🔐|🚀|Личный|Private/)
+mtpro.xyz##a[href^="/vpn.php"]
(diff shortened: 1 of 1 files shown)


Alex-302 (approved):

xglossyntx:
@Alex-302 you could have left `##a[href^="/vpn"]`. I didn't notice that one getting detected.

xglossyntx:
@ntnguyen1234 does this rule work for you without getting detected?
`#$#a[href][onclick*="this.href="][target="_blank"] { position: absolute !important; left: -3000px !important; }`

ntnguyen1234:
Yes, it works, but I don't think it will last long.

@Alex-302 Do you want to add?

xglossyntx:
The VPN one, however, could be re-added. it was there before without problems.

xglossyntx:
I'm not sure why this https://github.com/AdguardTeam/AdguardFilters/pull/228232#issuecomment-4241133883 rule works. I made one like it previously and they added detection for it, but now it seems to work again.

xglossyntx:
Could also just do something like this. It leaves a small leftover, but that's good because that's what they're checking for.
```
#%#//scriptlet('remove-node-text', '#text', '🚀 GET FAST VPN (VLESS)')
#%#//scriptlet('remove-node-text', '#text', '🔐 PRIVATE TELEGRAM PROXY')
```
It doesn't seem to remove the first one when you're on the proxy page.

PR #231199, opened 2026-05-13
Prepare files for AdGuard Mail Tracking Protection filter

Alex-302:
Related to https://github.com/AdguardTeam/FiltersRegistry/issues/1193
Rules will be moved to files later.

### DIFF
--- a/MailTrackingFilter/sections/allowlist.txt
+++ b/MailTrackingFilter/sections/allowlist.txt
@@ -1,0 +1,9 @@
+!
+! This section contains the list of rules that fix incorrect blocking.
+! Rules should be domain-specific.
+! If a rule causes a problem only in a specific email client, it should be restricted to that domain.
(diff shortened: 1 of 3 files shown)


zloyden (approved):

AdamWr (approved):

ntnguyen1234:
> ! This section contains the list of generic rules that are applied on every web-based mail client.

I think that "general" and "specific" should be differed whether the patterns are used in different websites, not in different mail clients?

For example `pixel.gif$image` can be used by different websites but its pattern will appear the same in gmail or yahoo mail.

Alex-302:
@ntnguyen1234 Updated. Is that clearer?

ntnguyen1234:
Yes, I think it's clearer now.

ntnguyen1234 (approved):

zloyden (approved):

